PR #79, opened 2015-01-14
BrowseSplitsIODialog: Merge redundant checks

lioncash:
'is' will be false if Tag is null.


CryZe:
Oh yeah, good find :+1:

PR #96, opened 2015-02-05
README update for compile issues

ajfischer:
I was having issues with getting LiveSplit to compile and couldn't really find anything online. I wanted to update the README with issues I had so other people don't have the same problem.

Additionally, is XSplit really a dependency in order to have VHMediaCOM installed and LiveSplit work, or am I doing something wrong?


CryZe:
Yeah, we implemented a XSplit Plugin, but never released it, and since this was before we used git, this is not even on a different branch. I think I'll remove it eventually / move it to a different branch.

PR #102, opened 2015-02-15
Undefine WITH_XSPLIT for now.

fatalis:
The first commit left xplit compiling enabled by default, but it looks like it should be disabled by default since it appears to be WIP code.


CryZe:
Yeah, it should be in it's own branch until we are done with it (this was before we used git). But this is fine too.

PR #118, opened 2015-02-24
Added autosplitters for Kotor 1 & 2

glasnonck:
I hope the game title is not case-sensitive.


CryZe:
I think it is. That should be fixed immediately.

PR #150, opened 2015-04-07
Fix LoadFactory attempting to load non-component DLLs and spamming event...

fatalis:
... log.


fatalis:
Was thinking of checking if dll starts with "LiveSplit." but wasn't sure if it'd break things.

PR #155, opened 2015-04-15
New split comparison generator: Weighted

ROMaster2:
Uses a percentile function for each segment to balance out the PB.


ROMaster2:
Maybe if git was being responsive that would have HAPPENED AppVeyor.

PR #166, opened 2015-04-25
TripleDateTime: Remove unused variables

lioncash:


wooferzfg1:
they're "used" in commented out code, so maybe the variables should be commented out or the commented out code should be removed too?

lioncash:
Updated it to remove the commented out code.

PR #171, opened 2015-04-28

stdscatchemall:
Added an edited version of SourceSplits to allow for L4D2 Auto-splitting. This is simply a test version however, fatalis said it can work, but until I test, I can't tell him otherwise.

### DIFF
--- a/LiveSplit.AutoSplitters.xml
+++ b/LiveSplit.AutoSplitters.xml
@@ -494,4 +494,14 @@
     <Type>Script</Type>
     <Description>Auto Start/Reset/Splitting is available for emulator (By cheeseandcereal)</Description>
   </AutoSplitter>
+  <AutoSplitter>
+    <Games>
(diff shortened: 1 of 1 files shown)


CryZe:
You should be able to just add this in your components folder, add it to your layout, maybe configure it and then you should be able to test it. Also you might want to do a Pull Request of the changes into SourceSplit itself, idk if it makes sense to have 2 SourceSplits.

stdscatchemall:
Yeah. I asked fatalis and he never got back to me, so I just went for it. As for auto-splitting. When i try to turn on auto-splitting for L4D2, it doesn't allow me to. I can't get to the configuration screen.

CryZe:
You can just add it to your layout like other components. You don't need to press the Activate Button in the Run Editor for that, it will still work. All the Activate Button does is adding a component to your layout without it being in your layout, kind of.

PR #194, opened 2015-05-15
Add Oblivion Loadless Timer

TheFuncannon:


LiveSplit.AutoSplitters.xml
```diff
@@ -494,4 +494,17 @@
...
+    <Games>
+      <Game>The Elder Scrolls V: Oblivion</Game>
+      <Game>Oblivion</Game>
+      <Game>TESIV</Game>
+      <Game>The Elder Scrolls 4: Oblivion</Game>
+    </Games>
+    <URLs>
+      <URL>https://raw.githubusercontent.com/TheFuncannon/Loadless.Oblivion/master/Oblivion.asl</URL>
+    </URLs>
+    <Type>Component</Type>
```
CryZe: The Type should be Script, otherwise it will attempt to load it as a component, which won't work.

LiveSplit.AutoSplitters.xml
```diff
@@ -494,4 +494,17 @@
...
+    <Description>Auto Splitting and Load Removal are available for patch 1.0. (By TheFuncannon)</Description>
```
CryZe: You claim it does auto splitting here, even though I don't see any auto splitting in the script. If you plan on implementing auto splitting, you should probably remove the Auto Splitting from the description and add it back once you implemented it. Otherwise people will be confused.

PR #200, opened 2015-05-20
ShortNameExtensions: Prevent multiple re-enumerations of an IEnumerable

lioncash:
Due to how IEnumerable functions, iterating over these directly will actually call GetShortNames() twice for each collection, which is not necessary. Reifying them into lists prevents this.

### DIFF
--- a/LiveSplit/LiveSplit.Core/Model/ShortNameExtensions.cs
+++ b/LiveSplit/LiveSplit.Core/Model/ShortNameExtensions.cs
@@ -31,8 +31,8 @@ namespace LiveSplit.Model
                 var splits = name.Split(new[] { splitToken }, 2, StringSplitOptions.None);
                 var seriesTitle = splits[0];
                 var subTitle = splits[1];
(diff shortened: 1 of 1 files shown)


CryZe:
GetShortNames() always returns an Array though, so it wouldn't build them multiple times.

lioncash:
GetShortNames' return type is IEnumerable<string> [eg](https://github.com/lioncash/LiveSplit/blob/enum/LiveSplit/LiveSplit.Core/Model/ShortNameExtensions.cs#L68)

CryZe:
So? It's still an Array and behave like one. The Array is not tied to the method anymore.

CryZe:
I merged it due to this still being a bit cleaner and safer, but it definitely didn't iterate over it multiple times.

lioncash:
Ah yes, this is true, I misinterpreted IEnumerable functionality. This would have applied to IEnumerables that are constructed via LINQ queries.

Issue #121, opened 2015-02-24
Fix Twitter OAuth


CryZe:
It might be broken. It didn't work for Jamacanbacn and me.

wooferzfg1:
https://github.com/CryZe/LiveSplit/commit/9d62a2a733ae1c9f0fcf8e4d35ac60c092563e33

PR #212, opened 2015-06-01
Twitter authorization modification

Seldszar:
Twitter share now use PIN authentication and fix Twitter sharing issue (#121).


CryZe:
This was already fixed in master. I don't see a point in deleting the whole OAuth Form :confused:

CryZe:
Wait what, looks like it was not fixed on master. Maybe some other branch? I remember fixing it.

wooferzfg1:
I never saw you fix it; I only saw you make the issue

CryZe:
I just fixed it: https://github.com/CryZe/LiveSplit/commit/9d62a2a733ae1c9f0fcf8e4d35ac60c092563e33

CryZe:
Sorry that you put quite some effort in this, but the solution was way simpler than deleting pretty much everything. I liked a few clean ups you did, so I guess you can do a Pull Request with those if you want. I don't want to get rid of the OAuth Form itself though, since having to copy paste a PIN around is just way worse for the user.

PR #223, opened 2015-06-06
Add Dredd to AutoSplitters XML

drtchops:
Add JDDvD ASL script and do other minor edits


wooferzfg1:
You had the type set as "Component" when it's an .asl file, so watch out for that next time :p

drtchops:
Whoops, I'll try to remember that for next time.

PR #241, opened 2015-06-16

VeritasDL:
added SADX autosplitter


wooferzfg1:
For next time:

LiveSplit automatically makes game names lowercase so capitalization
doesn't matter (SADX and sadx are identical) and we can't add two games
with the same key, so I removed "sadx" from the XML.

CryZe:
Gotta love the unit tests :D

wooferzfg1:
Yeah, it's very cool. There's no way I would've noticed that something was wrong otherwise

VeritasDL:
Okay thanks,

PR #243, opened 2015-06-16
Categorize the Recently Used List into Games and Categories

### LINKED ISSUE #240: Categorize the Recently Used List into Games and Categories
By Categorizing the Recently Used List we can have a lot more games there, it's a lot cleaner and you can easily find your splits.

CryZe:
By Categorizing the Recently Used List we can have a lot more games there, it's a lot cleaner and you can easily find your splits.

- [x] Categorize the Split Files
- [x] Simplify the Open Splits section
- [x] Add proper Backwards Compatibility
- [X] Order by Last Recently Used
- [X] Edit History Window

Fixes #240


wooferzfg1:
I don't think it's bad to categorize with very few items because it's more organized (even if it requires an extra click), but I'm not sure.

PR #251, opened 2015-06-19
Made cancel button affect Auto Splitters as well

### LINKED ISSUE #190: Autosplitters SetSettings(null) call in the splits editor
How to reproduce:
- Have splits set to a game that has an autosplitter with settings, make sure the autosplitter is deactivated before doing the next steps.
- Close the splits editor if it was open
- Open the splits editor
- Click Activate, don't open the settings
- Click Cancel

wooferzfg1:
The code we had before didn't do anything because it used
CurrentState.Run instead of runCopy and Auto Splitters weren't cloned so
the old settings could be set on an Auto Splitter without a component.

Currently the Auto Splitter's Component is disposed immediately when it is deactivated, but it should only be disposed if the user doesn't hit cancel (can be done similarly to Layout Components)

Fixes #190

### DIFF
--- a/LiveSplit/LiveSplit.Core/Model/AutoSplitter.cs
+++ b/LiveSplit/LiveSplit.Core/Model/AutoSplitter.cs
@@ -13,7 +13,7 @@ namespace LiveSplit.Model
     {
         Component, Script
     }
-    public class AutoSplitter
+    public class AutoSplitter : ICloneable
     {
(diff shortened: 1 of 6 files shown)


wooferzfg1:
The dispose issue is now fixed. Now this probably needs testing but everything seems to work as intended at the moment.

PR #253, opened 2015-06-19
Adding Box Out! Auto Splitting support

madzinah:


wooferzfg1:
It's a .asl so it should have type Script instead of Component.

madzinah:
Fixed this, sorry.

PR #255, opened 2015-06-20
Sound Component - Volume Mixer dependency

Seldszar:
Adding of the NAudio dependency required by the new Sound Component.


wooferzfg1:
Very nicely done.

PR #274, opened 2015-07-03
Core: Add missing SuppressFinalizer calls to LowLevelKeyboardHook

lioncash:
The destructor/finalizer shouldn't be executed on normal Dispose functionality.


CryZe:
You only need to suppress the finalizers if there's a finalizer. Since there's none, there's no need to call that.

lioncash:
I've amended the commit.

CryZe:
Okay cool. Looks good to me.

PR #275, opened 2015-07-04
Hotkey to Reset & Start

lietu:
One annoyance about LiveSplit was the inability with one key reset the timer to 0 and start a new run. When e.g. doing runs in Nuclear Throne, or any kind of permadeath game one thing you really care about is how to reset from scratch. At least in Nuclear Throne the deaths can happen rather frequently, so it quickly becomes an annoyance if you have to press more than one key to restart the game from scratch, and reset + start again the timer from 0.

This patch fixes this issue, and I'm able to set the "Reset & Start" -hotkey to R while playing Nuclear Throne. When I die, I press R to both 1) reset and start the timer from 0, 2) start a new run.

resetandstart

### DIFF
--- a/LiveSplit/LiveSplit.Core/Model/DoubleTapPrevention.cs
+++ b/LiveSplit/LiveSplit.Core/Model/DoubleTapPrevention.cs
@@ -105,6 +105,15 @@ namespace LiveSplit.Model
             Reset(true);
         }
 
+        public void ResetAndStart()
+        {
+            if (CheckDoubleTap())
+            {
+                InternalModel.Reset(true);
+                InternalModel.Start();
+            }
+        }
+
         public void Reset(bool updateSplits = true)
         {
             if (CheckDoubleTap())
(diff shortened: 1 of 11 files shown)


CryZe:
While I think this is a cool addition, I don't think it should be the second hotkey in the settings window. It's one of those hotkeys that almost no one will use, which is why I don't think it should stand out that much. So I feel hesitant to merge this before figuring out a clean solution for this.

lietu:
After looking into the code a bit, there is an alternative method I found for getting a similar effect with much less changes too.

Moving the hotkey code for the reset button from an `else if` down the list to a separate `if()` at the top, before `if (Settings.SplitKey == e)` so there's, `if`, `if`, `else if`, `else if`, and so on .. what this allows is that I can bind both "Start / Split" and "Reset" to the same key to achieve a similar effect.

https://github.com/LiveSplit/LiveSplit/blob/master/LiveSplit/LiveSplit.View/View/TimerForm.cs#L1013-L1016

Unfortunately this would require disabling Double Tap Prevention, and I think it would need a workaround for that.

wooferzfg1:
Wouldn't setting both Start / Split and Reset to the same key prevent you from splitting?

CryZe:
Having the same key for both would work for resetting and the splitting, but then there's no way of telling it to just split.

lietu:
True, it would prevent splitting .. can't quickly think of other options though.

CryZe:
Maybe have it in some kind of "Advanced" section that you can hide? I guess switching Comparisons could fit in there too

PR #277, opened 2015-07-04
Update PercentileComparisonGenerator.cs

ROMaster2:
Allows estimates that are added into the PB spot to be included when there's no splits recorded for that segment.

Okay that description is hard to understand, just look at the code lol.


ROMaster2:
wait, I just noticed something, need to fix

ROMaster2:
nevermind, bug only appeared with zfg's lobotomized split file. It should be good to go.

PR #278, opened 2015-07-05
Update PercentileComparisonGenerator.cs

ROMaster2:
It now can recognize when a segment is empty but there are still splits that can be used after it. This may be useful for the Average comparison as well.

Don't add immediately, may be really buggy. My tests came up good but it's better to be safe than let it break things right now.


ROMaster2:
Did more testing, appears to be stable. Feel free to merge.

PR #291, opened 2015-07-10
Fix the Window Titles so that OBS can capture it

### LINKED ISSUE #287: Make the Window Class Name of TimerForm uniquely identifiable
The Window Class Name or the Title of the TimerForm should be uniquely identifiable so that OBS can easily capture multiple LiveSplit windows without any problem. At the moment both the Window Class and the Title are always the same, so OBS has no way of distinguishing between both instances of LiveSplit.

CryZe:
Due to LiveSplit always having the same class name and same window title, OBS couldn't differentiate between the different LiveSplit windows, which is why a lot of people preferred using WSplit as a secondary timer. This fixes that.

Fixes #287 

Note: I'm not sure if I like the solution of setting it every update. Otherwise we need to use a lot of events though, which might not be that cool.

### DIFF
--- a/LiveSplit/LiveSplit.View/View/TimerForm.cs
+++ b/LiveSplit/LiveSplit.View/View/TimerForm.cs
@@ -271,14 +271,13 @@ namespace LiveSplit.View
         {
             var lowestAvailableNumber = 0;
             var currentName = "LiveSplit";
-            foreach (var process in Process.GetProcesses().Where(x => x.ProcessName == "LiveSplit").OrderBy(x => x.MainWindowTitle))
+            var processNames = Process.GetProcessesByName("LiveSplit").Select(x => x.MainWindowTitle);
+
+            while (processNames.Contains(currentName))
             {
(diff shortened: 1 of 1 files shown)


wooferzfg1:
I think the Layout name should be put before the Game & Category names since people are more likely to run the same game & category with two different layouts than the other way around.

wooferzfg1:
There's actually a few problems with this solution:
1) It doesn't solve the issue with multiple LiveSplit instances using Timer Only mode, since the Layout and Game Name/Category are blank.
2) If everything is blank, it displays "LiveSplit - - -" which looks pretty ugly.
3) If changing the Layout or Splits, OBS will no longer recognize LiveSplit anymore because the window title is different, so you'd have to re-add the Window Capture each time.

In my opinion, the only good solution would be if we displayed LiveSplit, LiveSplit (2), LiveSplit (3), etc. depending on how many instances of LiveSplit are open, but I'm not sure if that's even possible.

PR #292, opened 2015-07-10
Added a couple game names for Bioshock Infinite

Dread2472:


CryZe:
LiveSplit 1.6 is case insensitive, while 1.5 isn't. I advise against adding special cases for the casing. You should probably just wait for 1.6, which we hope to release soon anyway.

Dread2472:
Well can you change them to the upper-case S versions, then

CryZe:
You don't need to close Pull Requests to edit them btw.

wooferzfg1:
https://github.com/LiveSplit/LiveSplit/commit/fafed6af9e7470108d7eb962be8babc9183f3711 fixed

PR #302, opened 2015-07-13
Updated JKJA with Autosplitting

Dread2472:


wooferzfg1:
You can't have an & in an xml, though you can use &amp; instead. You should probably just change it to "and".

PR #323, opened 2015-07-21
Japanese translation

0xwas:


LiveSplit/LiveSplit.View/View/TimerForm.cs
```diff
@@ -2514,5 +2514,15 @@ private void RightClickMenu_Opening(object sender, CancelEventArgs e)
             RebuildControlMenu();
             RebuildComparisonsMenu();
         }
+
+        private void controlMenuItem_Click(object sender, EventArgs e)
```
CryZe: Unnecessary empty Events

LiveSplit/LiveSplit.View/View/AboutBox.Designer.cs
```diff
@@ -61,22 +61,22 @@ private void InitializeComponent()
             this.tableLayoutPanel.Controls.Add(this.lblVersion, 0, 1);
             this.tableLayoutPanel.Controls.Add(this.logoPictureBox, 1, 1);
             this.tableLayoutPanel.Dock = System.Windows.Forms.DockStyle.Fill;
-            this.tableLayoutPanel.Location = new System.Drawing.Point(9, 9);
+            this.tableLayoutPanel.Location = new System.Drawing.Point(9, 8);
```
CryZe: All of these seem to be done by accident.

CryZe:
Submodule URLs are unchanged, which is why AppVeyor can't find the commits, but it doesn't matter once we merged the submodules. Changing them would mean that we'd have to change them back after merging the Pull Request, so both solutions are kind of ugly.

CryZe:
Submodule Pull Requests are available here for example: https://github.com/LiveSplit/LiveSplit.Subsplits/pull/2

PR #336, opened 2015-07-31
Auto-Splitters.md: Fix a markdown link

lioncash:


CryZe:
gg, looks like my Atom Plugin is not that great (I already needed to make 2 changes to it, to make it any decent.

PR #338, opened 2015-08-01
Auto-Splitters.md: Use a table for the recognized types

lioncash:
Should be more helpful to those not familiar with C-like languages (or C#'s data types), or people who are taking their first shot at programming to attempt to make a script.


CryZe:
Good idea

PR #340, opened 2015-08-02
Program: Display the crash dialog in Release mode

lioncash:
Makes crash reporting from users easier, as they can now indicate (and moreover, actually see) what the reason for a crash is.

### DIFF
--- a/LiveSplit/LiveSplit/Program.cs
+++ b/LiveSplit/LiveSplit/Program.cs
@@ -39,15 +39,10 @@ namespace LiveSplit
                 if (Twitch.Instance != null)
                     Twitch.Instance.CloseAllChatConnections();
             }
-#if !DEBUG
             catch (Exception e)
             {
                 Log.Error(e);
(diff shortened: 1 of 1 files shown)


LiveSplit/LiveSplit/Program.cs
```diff
@@ -39,15 +39,10 @@ static void Main(string[] args)
                 if (Twitch.Instance != null)
                     Twitch.Instance.CloseAllChatConnections();
             }
-#if !DEBUG
```
wooferzfg1: This was !DEBUG (so Release only), which is what we want so that Visual Studio shows the error in Debug mode.

lioncash: tfw I can't read

lioncash: I'll update it to simply include the exception message, however, gimme a sec.

wooferzfg1:
We definitely want to display what the exception is, though. That was a cool change.

PR #345, opened 2015-08-04
TimerForm: Dispose of dialogs when done using them

lioncash:
btw SaveLayoutAs and SaveSplitsAs don't set the IsInDialogMode property like OpenSplits does. Not sure if there's a reason for that or not, so I left it untouched.


wooferzfg1:
Yeah, seems like IsInDialogMode needs to be set on those, good catch. I'll fix that.

PR #353, opened 2015-08-12
Removed Dark Souls Load Remover

Dread2472:
Only merge this if he adds the new dark souls IGT plugin


wooferzfg1:
https://github.com/LiveSplit/LiveSplit/pull/354 just replaced the AutoSplitter rather than adding one, so this pull request isn't needed.

PR #370, opened 2015-08-16
Update Auto-Splitters.md

fatalis:
byte400 Data : "", 0x000D6A1C, 0x11a570; // this actually works in the current version of ASL and it avoids needless repetition of module name


CryZe:
Wow, I didn't even know that and I wrote it xD

PR #375, opened 2015-08-17
List all contributors in the AboutBox

### LINKED ISSUE #346: Change contributors in AboutBox and website
As an open source project, there are a lot more contributors now other than CryZe and I, so we should maybe find a way to include everyone. Also, the donate button seems kind of sketchy as well, but I guess that doesn't matter much =p

wooferzfg1:
The list of contributors links to the contributors page on GitHub. The
donate button is also removed.

image

Resolves #346


CryZe:
Why was this closed? wontfix?

wooferzfg1:
yeah

PR #383, opened 2015-08-21
Update README.md

fatalis:
VS2010 doesn't support C# 5


CryZe:
We want to use C# 6 soon anyway, so idk if this change is necessary, as we'll have to change it again anyway.

fatalis:
Can we bump framework to 4.5 at that time too?  async/await is really powerful

CryZe:
Yeah, I guess

PR #395, opened 2015-08-24
Implement AutoCompletion for Game Names

### LINKED ISSUE #286: Improve Auto Completion for Game Names
This is really important since we definitely want the Game Names to be written correctly. Finding Zelda Games for example is really painful as the current auto completion logic requires you to fully write out the name from the left, so you definitely need to type at least "The Legend of Ze.." to even get to the Zelda games, which is super painful. A lot of people just won't do this, which is why the abbreviation algorithm + the fuzzy algorithm I implemented should be used there instead. Unfortunately implementing your own algorithm might not even be possible, so I hope we can still do this somehow.

wooferzfg1:
This uses the automatic abbreviations for game names in order to
generate suggestions in the Splits Editor.

Resolves #286


CryZe:
Clicking away should close the Drop Down Menu and pressing down should probably just choose the first element of the list (which we can do).

PR #398, opened 2015-08-25
Add Resident Evil 4 to ASLs (by LetsGetPitted)

drtchops:
Pitted did all the work, I'm just hosting the file for him.

### DIFF
--- a/LiveSplit.AutoSplitters.xml
+++ b/LiveSplit.AutoSplitters.xml
@@ -374,6 +374,19 @@
   </AutoSplitter>
   <AutoSplitter>
     <Games>
+      <Game>Resident Evil 4</Game>
+      <Game>Biohazard 4</Game>
+      <Game>RE4</Game>
+      <Game>re4</Game>
+    </Games>
+    <URLs>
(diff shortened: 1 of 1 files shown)


wooferzfg1:
Apparently one of these games already has an Auto Splitter?

drtchops:
5 and REmake do, this is for 4.

wooferzfg1:
Oh, nvm, the problem is that you can't add both RE4 and re4. LiveSplit isn't case sensitive with the Auto Splitter game names in 1.6.

drtchops:
Oh, good.

LiveSplit.AutoSplitters.xml
```diff
@@ -374,6 +374,19 @@
   </AutoSplitter>
   <AutoSplitter>
     <Games>
+      <Game>Resident Evil 4</Game>
+      <Game>Biohazard 4</Game>
+      <Game>RE4</Game>
+      <Game>re4</Game>
```
wooferzfg1: This needs to be fixed though. It should have either RE4 or re4, not both.

PR #402, opened 2015-08-25
View: Mark forms as localizable

lioncash:
Should aid with translation efforts in the future if anyone feels like localizing LiveSplit.

Most of this is just VS moving the UI values into the default language resxs. Behavior is the same as before, forms just have the Localizable property set to true.

### DIFF
--- a/LiveSplit/LiveSplit.View/View/AboutBox.Designer.cs
+++ b/LiveSplit/LiveSplit.View/View/AboutBox.Designer.cs
@@ -47,9 +47,7 @@
             // 
             // tableLayoutPanel
             // 
-            this.tableLayoutPanel.ColumnCount = 2;
-            this.tableLayoutPanel.ColumnStyles.Add(new System.Windows.Forms.ColumnStyle(System.Windows.Forms.SizeType.Percent, 100F));
-            this.tableLayoutPanel.ColumnStyles.Add(new System.Windows.Forms.ColumnStyle(System.Windows.Forms.SizeType.Absolute, 127F));
+            resources.ApplyResources(this.tableLayoutPanel, "tableLayoutPanel");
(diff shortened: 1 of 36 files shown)


wooferzfg1:
Not sure if this will conflict with https://github.com/LiveSplit/LiveSplit/pull/395 when we merge it, so I'll probably wait until that's merged before merging this.

lioncash:
Resolved conflicts from #395

lioncash:
Is there anything holding this back from being merged?

wooferzfg1:
We're not sure yet if we want Forms to be localized or just all the Labels' individual Text properties.

lioncash:
Having the forms localizable makes it much easier to manage. Each form has it's own resx namespace used (which is way better than managing it all manually). Also it lets you store exception messages and other code-side error messages in said resxes - which makes it less of a pain in the ass to translate/localize error handling strings.

Inevitably, if someone wants to translate a dialog into their own language, some UI elements will have to be shifted around to fit the theme/text-sizing regarding those languages (for example, Hebrew and Arabic dialects are read right-to-left, not left-to-right): 

Localizable resx's will store any positional changes made to the UI for a given language set (which can be selected through the Language property in the WinForms designer), so you have a clearer separation of UI layout settings.

wooferzfg1:
Wouldn't it be more difficult to manage if we make changes to a form though (like adding new settings)?

lioncash:
You can synchronize changes using https://visualstudiogallery.msdn.microsoft.com/3b64e04c-e8de-4b97-8358-06c73a97cc68 (VS will only update the culture invariant locale when doing UI changes). Normally, it would be up to the translator to keep their locale variant up to date.

wooferzfg1:
@CryZe thoughts?

lioncash:
Given this is 2+ years old, I'll close this

PR #403, opened 2015-08-25
Add simple overloads to DeepPointer

fatalis:
Return null as default for reference types instead of empty


fatalis:
appveyor will build fine when both this and the ASL request are merged

PR #407, opened 2015-08-29
Add MemoryWatcher - ASL "Current/Old" for DLL autosplitters

fatalis:
Rename len to numBytes in some string methods to avoid confusion


fatalis:
usage example [here](https://github.com/fatalis/LiveSplit.Dishonored/commit/073f7fc5d6749b9d750ae2be3b6d1733f1400b1f)

PR #420, opened 2015-09-06
Update Auto-Splitters.md

fatalis:
Good enough for now, will revise later.


CryZe:
Blaze it :smoking:

CryZe:
Dang, this looks really good. Good job :+1:

PR #428, opened 2015-09-12
Implement a Contextual Slideshow Component

### LINKED ISSUE #305: Implement a "Contextual Component"
This component would basically have its own layout editor and you can put other components in it. It only renders a single component at a time though and stores a queue of components that want to be rendered. Basically whenever a component invalidates, it gets stored in the queue and the next component in the queue is shown every 5 seconds or so, if there's one in the queue. Otherwise the current component stays active. A component can't be in the queue multiple times, otherwise other components would probably never be seen, if there's a component invalidating all the time.

This component would be really useful as you could combine multiple Info Text Components into a single one, meaning that it would take way less space, but also always shows the one that's currently relevant.

I guess you could also auto add components that don't invalidate a lot to the queue every x seconds.

CryZe:
Closes #305 

Progress:
- [x] Show the individual components
- [x] Switch between the components
- [x] Prioritize Components
- [x] Swap in Components every now and then that didn't invalidate for a long time
- [ ] Implement Settings
- [ ] Implement a Layout Editor
- [ ] Fix the Height Bug

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -85,3 +85,6 @@
 [submodule "LiveSplit/Components/LiveSplit.TotalPlaytime"]
 	path = LiveSplit/Components/LiveSplit.TotalPlaytime
 	url = https://github.com/LiveSplit/LiveSplit.TotalPlaytime.git
+[submodule "LiveSplit/Components/LiveSplit.ContextualSlideshow"]
+	path = LiveSplit/Components/LiveSplit.ContextualSlideshow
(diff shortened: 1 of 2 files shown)


wooferzfg:
We should probably embed a slightly modified Layout Editor into the Contextual Slideshow's settings, and this embedded Layout Editor should have a button to edit the inner components' settings.

CryZe:
We need that for Embedded Layout and Window Component too and maybe others, so I guess that should just be available in LiveSplit.Core or LiveSplit.View

PR #434, opened 2015-09-17
Added load remover for "Grand Theft Auto V"

zoton2:
Hopefully I didn't mess anything up this time.


fatalis:
just fyi, can be shortened to

```
isLoading
{
	if (version != "")
		return current.loading > 0;
}
```

zoton2:
Ah OK thanks; have just updated the script to work like that. I was too used to all the complicated stuff I was doing for the Vice City autosplitter that I forget how to keep things simple I guess.

PR #437, opened 2015-09-24
Add BioShock 2 Load Remover

Glurmo:


fatalis:
for future reference, you should do it like this

```
state("Bioshock2")
{
	bool isSaving : 0xF42EE8;
	bool isLoading : 0x10B8010, 0x278;
}
```

Glurmo:
Thanks, I updated the script to use the 1.6 style. I originally wrote it before 1.6 was released, didn't notice the recommended syntax changed.

PR #441, opened 2015-10-01
Use C#6 and clean up

CryZe:
This makes LiveSplit's code use most of C# 6's features like the Lambda Syntax for writing Properties and Methods. Also this cleans up a lot of LiveSplit's code. .NET 4.5 is being used from now on as well.


fatalis:
👌

CryZe:
Looks like I broke something with the Contextual Component.

CryZe:
Okay, rebased and fixed the Contextual Component bug (it wasn't even supposed to be part of this, as its not on master yet).

CryZe:
Reminder: Don't forget to fast-forward all the master branches of the submodules to the submodule HEADs.

wooferzfg:
The README should be changed too (C#6, only VS 2015, .NET 4.5)

PR #443, opened 2015-10-03
Random cleanup in Core

fatalis:


LiveSplit/LiveSplit.Core/Model/IRun.cs
```diff
@@ -164,21 +164,20 @@ private static void RemoveNullValues(IRun run, TimingMethod method)
 
         private static void RemoveDuplicates(IRun run, TimingMethod method)
         {
-            for (var index = 0; index < run.Count; index++)
+            foreach (var segment in run)
```
CryZe: Please revert this, because you can't modify the collection in a foreach loop.

PR #444, opened 2015-10-04
Fix issue in previous commit

fatalis:


wooferzfg:
What's the reason for this change?

fatalis:
Cryze said to revert it due to collection modification

wooferzfg:
Don't you only modify the segment's Segment History, not the segment/collection itself?

CryZe:
I misread the code. I'm sorry. There's no actual reason for this.

PR #460, opened 2015-12-04
Fixed Magnetic By Nature 404

0Seren:
Sorry about that...


wooferzfg:
type should be Script, not Component.

PR #461, opened 2015-12-04
Added load remover for "SAW"

zoton2:
Added a few aliases for what someone might type in for this game but hardly anyone is gonna run it I doubt so I'm not going to type in a lot of names.


CryZe:
ASL is not case sensistive. Therefore you should remove the duplicated entries.

wooferzfg:
Saw and SAW are considered the same game because LiveSplit isn't case sensitive, so you should only have one of those two in your list of games.

zoton2:
Yeah I thought that might be the case; removed.

PR #483, opened 2016-01-26
Write Split Informations in Text Files

ExploudYourEar:
A background Worker is writing informations in a text file.
There is a sefty code in case the program can't handle the start/stop
process because people are spamming Skip, Undo or Split for some
reasons.
Those informations are :
- Starting Time of an Attempt (local time)
- Ending Time of an Attempt (local time)
- Name of a split
- PB Segment TIme
- Best Segment Time
- Split Time
- Current Comparison
- Timing Method.
  The background worker let livesplit continue without any freeze while
writing informations in text files.


fatalis:
You should use the LivesSplit Server component to interface with LiveSplit.

PR #531, opened 2016-04-03
Installed font

Hathor86:
Use  InstalledFontCollection().Families instead of FontFamily.Families
to list all installed font

Also small performance improvement
Create a static array to hold FontStyles; avoinding recreation for each
font.

### DIFF
--- a/LiveSplit/Libs/CustomFontDialog/CustomFontDialog/FontList.cs
+++ b/LiveSplit/Libs/CustomFontDialog/CustomFontDialog/FontList.cs
@@ -2,28 +2,31 @@
 using System.Drawing;
 using System.Windows.Forms;
 using System.Runtime.InteropServices;
+using System.Drawing.Text;
 
 namespace CustomFontDialog
 {
     public partial class FontList : UserControl
     {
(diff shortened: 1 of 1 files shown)


wooferzfg:
what's the advantage of using InstalledFontCollection()?

Hathor86:
Following .Net doc, FontFamily.Families 
> Returns an array that contains all the FontFamily objects associated with the current graphics context.

and InstalledFontCollection
> Represents the fonts installed on the system. 

I installed some font on my PC and I quickly remark that some was missing. So I had a look on the internet and find that using InstalledFontCollection gets (as it is named) fonts installed.
I made a test and get what I expected but...

I was curious and I dissambled System.Drawing.dll and here what you found behinf FontFamily.Families property
> return new InstalledFontCollection().Families

So... Honestly, I don't really know what is the real difference between them (and what's happening in  the dark side of .Net) but the fact is that fonts I installed where listed in the second case and not in the orginal file.
(And I ssaid to myself that maybe other people can meet this problem, so I made this pull request :) )

wooferzfg:
Is there a specific reproducible case that your changes affect? I didn't notice any difference between the old and new behavior.

PR #541, opened 2016-05-17
Multi-row move

abney317:
Allows for all selected rows to be moved up or down at the same time in the Edit Splits dialog.


wooferzfg:
Moving a single row up once makes it also select the row that's 2 below it. Not sure why.

abney317:
Should be fixed now

PR #544, opened 2016-05-20

Grim995:
Added asl for Need for Speed: Underground 2


fatalis:
btw

`refreshRate = 500;`
will be forced to ~60

PR #545, opened 2016-05-20
Import comparison by index

abney317:
If same number of splits then it will just add by index, otherwise the name method will be used
Bit of a work around for the bug case that a game has multiple splits with the same names

### DIFF
--- a/LiveSplit/LiveSplit.View/Model/RunImporters/IRunImporter.cs
+++ b/LiveSplit/LiveSplit.View/Model/RunImporters/IRunImporter.cs
@@ -1,4 +1,4 @@
-﻿using LiveSplit.UI;
+using LiveSplit.UI;
 using System.Linq;
 using System.Windows.Forms;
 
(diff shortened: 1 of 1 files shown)


wooferzfg:
I'm not sure that this is the best solution, since you can have a comparison that is a different route with the same number of splits. It would probably be better to have something that handles the specific case of multiple splits that have the same name.

abney317:
I agree, there should be a better way of going about the name method. I still think defaulting to using the index is better at the moment at least because I imagine this is the most common case for people importing splits. Usually people would be importing splits from the same route I'd think. For example if I'm importing splits where someone used a full level's name as the split name but all of my splits use an abbreviation of the levels name, then trying to match the names isn't going to work at all.

wooferzfg:
I made a more general fix for the case of multiple splits having the same name: https://github.com/LiveSplit/LiveSplit/commit/0a1aaa4e1d767b44a8f8a8c4dbee8a4a909c6ced

PR #576, opened 2016-07-23
2 New ASLS with annoying names

Dread2472:


wooferzfg:
you have "Jedi Knight Mysteries of the Sith" twice

Dread2472:
Yep I really shouldn't have copy pasted, but to be fair the games have really dumb names

wooferzfg:
it's all good, i would do the same thing lol

Issue #581, opened 2016-08-02
Lack of User Understanding about Game Time vs Real Time


0Seren:
## Issue
Users do not seem to know about switching from Real Time to Game Time comparisons when they have an autosplitter enabled. And get confused as to why loads are still counted in their times or why the IGT isn't pausing or matching with the game's shown IGT.

## Proposed Solutions
- Add a warning when an autosplitter is activated, but LiveSplit is comparing against Real Time. Artist Rendition: ![autosplitter](https://cloud.githubusercontent.com/assets/13229561/17340571/ca7dc8e0-58be-11e6-8400-3e44af51a128.png)
- Automatically switch to comparing against Game Time once a user has clicked "Activate" in their splits editor, and notify the user via modal or text area, and explain how to revert back/why it was changed to Game Time
- Make Game Time act as Real Time when no autosplitters are found and default to Game Time

I listed the proposed solutions in the order I think is best. Other solutions/options are welcome.

SuiMachine:
How do I upvote stuff? xD

Dalet:
I think a MessageBox would be better for a warning. (#630)
Maybe show it only the first time the user activates an autosplitter.

A Messagebox could also ask if you want to switch to Game Time and do it for you. (similar to your second proposal)

If we want to show either of those only for autosplitters that use Game Time we could just attempt to detect it if it's ASL, and for the rest make it opt-in with an element in the autosplitters XML.

0Seren:
I personally dislike the MessageBox due to its intrusiveness and its one time nature. To fix the later it could simply have a checkbox that says "Remember my choice for all future auto splitters." or w/e. However, I do believe that instead of a MessageBox, keeping it part of the original interface would be better. Since game time is only important for No Loads or IGT detecting if either of those are implemented could be a way to do it. 

Personally I feel as though the xml is straightforward enough that it might be better to use a webform that implements the xml instead, which would make your opt-in element be more viable imo. However, that's a bit of a different discussion.

For now I agree it would be nice to have it only display anything when it applies to that auto splitter. I've done an **extremely** rough draft of what this might look like. I actually don't like how the buttons and text are laid out in this example, but it should give a rough idea of what I'm talking about.
![compare to game time sample](https://cloud.githubusercontent.com/assets/13229561/20488381/ca20171e-afd4-11e6-9400-7261ea17afd0.png)

PR #630, opened 2016-10-22
Added Comparison Warning For Auto Splitters

0Seren:
Added comparison warning for Auto Splitters when not comparing to Game Time. Based around issue #581 with hopefully better wording.

It makes it go from:
Before
To:
After


0Seren:
Is there an issue with this, or are you just not sure about adding it? Because if there's an issue or something you'd like changed I can work on fixing it. (i.e. show/hide warning, different text, etc.)

wooferzfg:
The main issue I have with it besides there being so much text is that some Auto Splitters don't use Game Time so it might be confusing to users if it still pops up in that case. And also I'm not sure if this is the best long term solution.

PR #672, opened 2017-01-12
Remove split key resets finished run

hallcristobal:


CryZe:
Check if it's timer only instead. I definitely want it to stay for that mode.

LiveSplit/LiveSplit.View/View/TimerForm.cs
```diff
@@ -919,7 +919,10 @@ private void StartOrSplit()
             }
             else if (CurrentState.CurrentPhase == TimerPhase.Ended)
             {
-                Model.Reset();
+                if (InTimerOnlyMode)
```
CryZe: You can collapse these into a single if

hallcristobal: Silly mistake, my bad

PR #683, opened 2017-01-20
Master

Lyelt:
License updated


wooferzfg:
This commit should only be on your own fork.

PR #727, opened 2017-03-07
Added Auto Splitting for Dark Souls III

pawREP:


pawREP:
Build check failed because I tried to add a script to a game that already has a "Component" extension. How do I properly submit a script in this case?

PR #757, opened 2017-05-01
Added INK to Auto-Splitter

ROMaster2:


CryZe:
I think it's case insensitive, that's why it failed

wooferzfg:
Game name aren't case sensitive, so you need to remove one of those.

PR #802, opened 2017-06-24
Fix crash when removing all segments in Run Editor

hallcristobal:


CryZe:
We may want to handle this differently and need to look into how LiveSplit One diverged and if it somehow applies there too.

PR #804, opened 2017-06-27
Trackmania² Lagoon Autosplitter 

RastaBobby:
https://www.dropbox.com/s/xv1bdy7hf6darji/Lagoon.asl?dl=0

PS: I don't know if this is the right place to send a new autosplitter lol


wooferzfg:
Make a new GitHub repo, upload your file to it, get the raw download link from there, add that link to the AutoSplitters XML (https://github.com/LiveSplit/LiveSplit/blob/master/Documentation/Auto-Splitters.md#adding-an-auto-splitter)

PR #805, opened 2017-06-27
Fixed crash when font size is set to zero in font dialog

miere43:
Fixes this thing:

test

Just clamps zero to one now.


wooferzfg:
Merged here: https://github.com/LiveSplit/LiveSplit/commit/90b362ba65a623ba3d618dd4ad8c9260f014fc60

PR #828, opened 2017-07-19
Adding an autosplitter for The End Is Nigh

Grimelios:
I've never used this file before, so I'm not sure if I'm storing my DLL in the correct location. I can easily move it to another folder if required.

### DIFF
--- a/LiveSplit.AutoSplitters.xml
+++ b/LiveSplit.AutoSplitters.xml
@@ -2947,4 +2947,16 @@
     <Description>auto starts / resets and splits at every screen.</Description>
     <Website>https://github.com/Berumondo/Volgarr-autosplitter</Website>
   </AutoSplitter>
+   <AutoSplitter>
+    <Games>
+      <Game>The End Is Nigh</Game>
+    </Games>
+    <URLs>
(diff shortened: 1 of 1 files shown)


wooferzfg:
I don't think you have the DLL stored in your repository, so you can either add it somewhere or just use your release link (https://github.com/Grimelios/LiveSplit.TheEndIsNigh/releases/download/v1.1.0/LiveSplit.TheEndIsNigh.dll). However, whenever you make a new release, you'd have to update the link in the AutoSplitters XML.

Grimelios:
Okay. I'll put the DLL in a folder within the repository so I won't have to update the Autosplitters XML with every new release. One additional question. How do updates (and specifically update notes) work when using the XML file? Ordinarily, I'd update my own Updates.xml file for the component and let LiveSplit update the component automatically on launch.

wooferzfg:
Here's an example update XML file: https://fatalis.pw/livesplit/update/Components/update.LiveSplit.SourceSplit.xml

After you add the XML file to your repository, make sure that the URL is correct in the Factory.

Grimelios:
Awesome. Thank you. I'll update the AutoSplitters XML file again soon.

PR #846, opened 2017-08-10

TheFuncannon:
Updated IGT with auto-start and reset for the current version of the game.


wooferzfg:
You need to change the type to Script instead of Component.

PR #882, opened 2017-10-01
Mouse passthrough when timer running.

zzymyn:
Added a layout option that makes mouse clicks pass-through when the timer is running, so you can't accidentally click the window when in-game. I've only tested it on Win10.

### DIFF
--- a/LiveSplit.AutoSplitters.xml
+++ b/LiveSplit.AutoSplitters.xml
@@ -3279,4 +3279,14 @@
     <Website>https://github.com/ShootMe/LiveSplit.SteamWorldDig2</Website>
     <ShowInLayoutEditor />
   </AutoSplitter>
+  <AutoSplitter>
+    <Games>
+      <Game>Cuphead</Game>
+    </Games>
+    <URLs>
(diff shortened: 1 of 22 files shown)


wooferzfg:
The main issues I have with this:

- Using this with Global Hotkeys off can be confusing, since it may seem like LiveSplit is completely unresponsive unless you Alt+Tab to get focus back.
- If you don't have Split or Reset assigned as hotkeys, LiveSplit becomes completely unusable as soon as you start a run.

I think the best solutions would be to either have a warning message for when these two cases come up or to have a way to gain mouse control back.

Also I'm not sure if this setting should be in the Layout or in the Settings. "Always on Top" is in a weird gray area as well.

zzymyn:
Made a small change so that the mouse pass-through is only enabled when the timer window is in the background (logically not visually). Means you can still interact with it if you alt-tab to it or select it from the task-bar.

wooferzfg:
I think adding "and Not In Focus" might be better. Thoughts?

livesplit vshost_2017-10-02_19-31-51

zzymyn:
Yeah I think that makes sense.

Issue #416, opened 2015-09-02
Overflow Exceptions in WndProc


wooferzfg:
https://twitter.com/HarutomoSRL/status/638882690339483649

wooferzfg1:
This is definitely an issue with negative screen coordinates, but it seems like we use the proper implementation of the Point constructor, so I have no idea.

wooferzfg:
Duplicate of #2328

PR #885, opened 2017-10-02
IntPtr.ToInt32 will throw an OverflowException if it's < 0.

zzymyn:
#416

### DIFF
--- a/LiveSplit/LiveSplit.View/View/TimerForm.cs
+++ b/LiveSplit/LiveSplit.View/View/TimerForm.cs
@@ -1443,7 +1443,7 @@ namespace LiveSplit.View
             if (m.Msg == WM_NCHITTEST || m.Msg == WM_MOUSEMOVE)
             {
                 Size formSize = Size;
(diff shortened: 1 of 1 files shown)


zzymyn:
See https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/IntPtr.cs#L115 for IntPtr.ToInt32 code.

wooferzfg:
It's a signed integer, though. IntPtr.ToInt32 should only throw an exception if the number's magnitude is too large to be represented with 32 bits.

zzymyn:
Hmm yes that's true. I could have sworn I'd had this happen but can't get it to happen again even with the original code now...

wooferzfg:
Does it maybe have to do with the OS?

zzymyn:
Maybe running on a 32-bit Windows? Expansion of the lparam from 32 bits to 64 bits?

wooferzfg:
The person in the tweet in #416 said they were on 64-bit though...

zzymyn:
Right. I will investigate further.

PR #888, opened 2017-10-05
added Quantum Conundrum load remover

blairmadison11:
I have created a load remover for Quantum Conundrum. Tested and working for latest Steam version. Will try to add start/split/stop in the future.


blairmadison11:
whoops typo'd it. will resubmit the request

PR #891, opened 2017-10-07
Use CredentialManager to store passwords and access tokens

### LINKED ISSUE #883: Get rid of WPF, XSplitPlugin, OBS Plugin
Depends on if we ever plan to maintain/develop any of these.

### LINKED ISSUE #887: Use Windows Credential API for storing passwords and Login Forms
https://gist.github.com/judwhite/15b48759264efa97d2da413600184c3e

wooferzfg:
This seems to work well for SRL. However, we should consider adding a way to clear access tokens for the other sites (Twitch, Twitter, Speedrun.com) because there is no way to log out and log in as a different account at the moment.

Resolves #887 
Resolves #883


hallcristobal:
I think just a simple `Logout` or `Clear Credential(s)` button would work fine as far as "log out and log in as a different account". Logging out should just clear the credentials using `CredentialManager.DeleteCredential(...)` if I'm understanding this correctly.
Also, should probably add the license file for attribution [https://github.com/meziantou/Meziantou.Framework](https://github.com/meziantou/Meziantou.Framework).

PR #895, opened 2017-10-09
Fix Saving Images

### LINKED ISSUE #894: Certain split files fail to save
Some images can't be saved properly.

[example_file.zip](https://github.com/LiveSplit/LiveSplit/files/1369641/example_file.zip)

CryZe:
Image objects borrow from MemoryStreams. If they are closed and the
Image objects are saved again, then a generic GDI+ error is caused.

Fixes #894


wooferzfg:
Does this need to be changed as well?

https://github.com/CryZe/LiveSplit/blob/02b4866ec7deb26ebc44ad24d3928bf1a9f0f7fb/LiveSplit/LiveSplit.Core/UI/SettingsHelper.cs#L128

CryZe:
Considering it always worked, I don't think so. I bet bf.Deserialize works differently.

PR #924, opened 2017-11-13
Wolfenstein II: The New Colossus

itsjabo:
Okay finally..


wooferzfg:
You didn't close your AutoSplitter tag. You might also need to change the file type to be .asl, but I'm not 100% sure on that.

itsjabo:
How do i close the autosplitter tag?

PR #928, opened 2017-11-13
Load remover wasnt working

itsjabo:
Trying new url, seems to me that my old URL seemed different to the majority of URLS on here, so i changed mine to see if it fixes the problem of the load remover not being found by LiveSplit on the edit splits section.

### DIFF
--- a/LiveSplit.AutoSplitters.xml
+++ b/LiveSplit.AutoSplitters.xml
@@ -3508,7 +3508,7 @@
       <Game>Wolfenstein II: The New Colossus</Game>
     </Games>
     <URLs>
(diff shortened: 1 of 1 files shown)


wooferzfg:
It shows up for me though?
livesplit_2017-11-13_17-48-54

wooferzfg:
Oh you meant that Activation fails, nvm

wooferzfg:
I think the problem might be that you have spaces in the name of the .asl file itself, so after it downloads, it can't save properly.

itsjabo:
Yeah activating it doesnt work "the auto splitter could not be activated. (the given path's format is not supported.)

wooferzfg:
Anyways I'll merge and we'll see if just changing the link fixes it.

PR #943, opened 2017-12-02
Add AutoSplitter for INFRA

gurka:
Load Removal only.


gurka:
INFRA is built using the source game engine, so I guess that it would actually be better to add support for INFRA to SourceSplit instead of directly into LiveSplit. See https://github.com/fatalis/SourceSplit/issues/12 No response in 1½ month, so while we wait for SourceSplit to add support I think it would be good to add this basic support (Load Removal only) directly to LiveSplit. Credits for the script to @BigJim123

PR #953, opened 2017-12-18
Resume Run Functionality 

Kigoli:
Enhanced LiveSplit to be able to recreate a run where it was last left off. 

Changes:

- LiveSplitCore.cs
   * Added code to read the StopTime from LiveSplit-Core.dll
- IRun.cs
-  Run.cs
  * Added the new StopTime variable. This variable saves what the timer read when the run was paused.
-  LiveSplitsState.cs
   * Added the new resumedRun variable. This variable saves if current run was resumed from a save file.
   * Initializing the new variable to false.
   * Added the ResumeSplits function to rebuild where we left off, as well as setting the currentState in the way that we need it for proper functionality.
-  StandardFormatsRunFactory.cs
   * Parsing the StopTime.
-  JSONRunSaver.cs
-  WSplitRunSaver.cs
-  XMLRunSaver.cs
   * Added code to write the new variable to the save files.
-  TimerModel.cs
   * Updated the Start function to call the Pause function if we're a resumedRun.
   * Skipping the call to UpdateAttemptHistory if we're on a resumed run. Treating it as a continuation and not a brand new run.
   * Resetting the resumedRun flag when we Reset().
   * While we're updating the segments, we need to remove an existing entry for the current attempt. This shouldn't happen, but if it does we'll default to the new value.
-  livesplit_core.dll
   * Updated this to parse the new StopTime variable.
-  TimerForm.Designer.cs
   * Added code for the new "Resume Run..." option.
-  TimerForm.cs
   * In SetRun, call ResumeSplits if we're from the Resume Run option.
   * More code to support the new Resume Run option.

### DIFF
--- a/LiveSplit/LiveSplit.Core/LiveSplitCore.cs
+++ b/LiveSplit/LiveSplit.Core/LiveSplitCore.cs
@@ -2027,7 +2027,7 @@ namespace LiveSplitCore
             {
                 throw new ObjectDisposedException("this");
             }
-            var result = new TimeSpanRef(LiveSplitCoreNative.Run_StopTime(this.ptr));
+            var result = new TimeSpanRef(LiveSplitCoreNative.Run_stop_time(this.ptr));
             return result;
         }
         public TimeSpanRef Offset()
@@ -5127,7 +5127,7 @@ namespace LiveSplitCore
         [DllImport("livesplit_core", CallingConvention = CallingConvention.Cdecl)]
         public static extern IntPtr Run_metadata(IntPtr self);
         [DllImport("livesplit_core", CallingConvention = CallingConvention.Cdecl)]
-        public static extern IntPtr Run_StopTime(IntPtr self);
+        public static extern IntPtr Run_stop_time(IntPtr self);
         [DllImport("livesplit_core", CallingConvention = CallingConvention.Cdecl)]
         public static extern IntPtr Run_offset(IntPtr self);
         [DllImport("livesplit_core", CallingConvention = CallingConvention.Cdecl)]
--- a/LiveSplit/LiveSplit.Core/LiveSplitCore.cs
+++ b/LiveSplit/LiveSplit.Core/LiveSplitCore.cs
@@ -2021,6 +2021,15 @@ namespace LiveSplitCore
             var result = new RunMetadataRef(LiveSplitCoreNative.Run_metadata(this.ptr));
             return result;
         }
+        public TimeSpanRef StopTime()
+        {
+            if (this.ptr == IntPtr.Zero)
+            {
+                throw new ObjectDisposedException("this");
+            }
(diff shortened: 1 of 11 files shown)


Kigoli:
@wooferzfg I see that my code failed the continuous-integration checks. 

What are the next steps? I fix the issues and then recommit? Any help would be appreciated.

wooferzfg:
I think it only failed because you made changes in LiveSplit.Timer, and it couldn't retrieve those changes. Maybe make a PR in LiveSplit.Timer with the changes that you made?

wooferzfg:
I'm not sure if it's even possible for CI to pass until the changes in LiveSplit.Timer get merged, but that's fine

CryZe:
How does this work specifically? It seems like only store a StopTime in the splits. Don't you lose all kinds of other information then?

Kigoli:
@wooferzfg Hmmm, okay, I'll try that.

@CryZe Well, I only worked on this for a few days, so I may have missed some crucial pieces. 

A few assumptions I made:

 * The Save/Load functionality does not capture the time that the Timer was paused at.
 * The pieces we need to recreate a run are: AttemptNumber, Split Times, and Paused Time.
 * There's nothing critical to a run that we aren't saving/loading currently.

===================================================
So, the Run object we create from from LoadRunFromFile has a list of Segments, and each Segment has a SegmentHistory which records the time of each split for each attempt. 
Run
  - List <Segment> Segments
    - List <Segment> SegmentHistory
       - Index ( This is which Attempt this Split was part of. )
       - Time ( This is the Split Time, or rather how much time the Split Took. ) 
===================================================
When we resume a run, we grab the Attempt Number. This is important, as when we save a run, the Attempt Number does not increase/decrease. 

Int AttemptNumber; ( This was the AttemptNumber that the saved run was part of. ) 
===================================================
We then loop through each Segment, and grab the Split time for that Segment for that AttemptNumber. 

For( each Segment in Run )
    Time SplitTime = Run.Segments.SegmentHistory.at( AttemptHistory );
===================================================
We can then "recreate" the run but "splitting" at the appropriate times.
    Time TotalTime = TotalTime + SplitTime;
    CurrentSplit.SplitTime = TotalTime;
    CurrentSplitIndex++;
===================================================
We need to set the TimePausedAt to the StopTime we got from the Save File. We also need to set the CurrentPhase to Paused, as that's what it really was before we Saved the run.
    TimePausedAt = Run.StopTime
    CurrentPhase = TimerPhase.Paused;
===================================================
Lastly, we need a way to tell that we're on a resumed run for control purposes.
    resumedRun = true;
===================================================

The rest is processing button presses slightly differently. 

So, are there any critical pieces you think I'm losing when I save/resume the run? Do my comments make sense? This is my first time working on an OpenSource project, so I'm kinda new to all of this.

MinecraftPlaye:
Any news on this one as a few friends of mine and myself are currently searching for a program which can pause and resume your run even after it was closed.

PR #965, opened 2018-01-01
Make ModulesWow64Safe notice reloaded modules

kugelrund:
Cached modules are now checked for validity. If a call to
'GetModuleInformation' for a module is not successful, the cache is
cleared. Modules that are reloaded during the lifespan of the process
can therefore be tracked correctly now.


kugelrund:
Needs benchmarking before merging.

kugelrund:
In this form unfortunately way to heavy on performance. 96% of the time in ModulesWow64Safe() was spent in IsModuleArrayValid() in a quick benchmark.

PR #992, opened 2018-01-24
Fix divide-by-zero causing weird scaling on startup.

qwertymodo:
With certain layout dimensions, LiveSplit will rescale the saved dimensions in weird ways on startup, such as this layout:

rightsize

looking like this instead once the program is closed and re-opened:

wrongsize

This seems to be due to a divide-by-zero error on startup, when OldSize == 0 in TimerForm.cs

### DIFF
--- a/LiveSplit/LiveSplit.View/View/TimerForm.cs
+++ b/LiveSplit/LiveSplit.View/View/TimerForm.cs
@@ -1194,11 +1194,14 @@ namespace LiveSplit.View
             {
                 if (OldSize != currentSize)
                 {
-                    MinimumSize = new Size(0, 0);
-                    if (Layout.Mode == LayoutMode.Vertical)
-                        Height = (int)((currentSize / (double)OldSize) * Height + 0.5);
-                    else
-                        Width = (int)((currentSize / (double)OldSize) * Width + 0.5);
+                    if (OldSize != 0.0)
+                    {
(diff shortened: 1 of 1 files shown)


wooferzfg:
Could you upload your layout file that has this issue?

qwertymodo:
Not without also uploading my component, which isn't currently publicly available.

wooferzfg:
You should probably have your custom component avoid ever returning a VerticalHeight/HorizontalWidth of 0. But also, the ComponentRenderer should probably prevent the overall size from being 0 when drawing.

qwertymodo:
This is still a good sanity check to prevent a size of 0 from actually propagating into a divisor.

wooferzfg:
I just made a change that should prevent the OverallSize (and in turn, OldSize) from ever being 0: https://github.com/LiveSplit/LiveSplit/commit/864e9b66ac9c55e2e352a99f5c0f25bf6343350f

qwertymodo:
OverallSize wasn't the problem, it was set to 1024 just fine.  It was only OldSize that was initialized to 0, resulting in a division-by-zero the very first time that currentSize was nonzero.

qwertymodo:
Ok, I tried your patch and it resolved the symptoms.  If you'd rather go that way than with this PR, that works too.

wooferzfg:
Alright. Thanks for finding the issue!

PR #1042, opened 2018-04-06
Added Kao 2 autosplitter

RibShark:


wooferzfg:
Could you change the tab before the <AutoSplitter> tag so that it is two spaces instead?

RibShark:
Done. I don't know how that happened

PR #1075, opened 2018-05-06
Use 64bit ints for deeppointer

### LINKED ISSUE #1034: New DeepPointer absolute pointer constructor doesn't work for 64-bit games
In commit d48ea3f (from #969), a new constructor for DeepPointer was added that allows you to set an absolute address as the base, instead of having to set it relative to a module. This seems like a nice addition, but has a significant flaw: Because the address is passed as an `OffsetT`, which is defined as `Int32`, this will cause issues if you try to use an absolute address with a game using a 64-bit executable. I tried changing `OffsetT` to `Int64` to test it, but it caused issues with the `IntPtr` operations.

This constructor should really take an `IntPtr` in its constructor. This change hasn't made it into any release, so it *might* be safe to replace the constructor, but it could also be added as a separate constructor if removing the constructor isn't a practical solution for the project.

res0nance:
This should fix #1034 along with another PR in 
https://github.com/res0nance/LiveSplit.ScriptableAutoSplit

### DIFF
--- a/LiveSplit/LiveSplit.Core/ComponentUtil/DeepPointer.cs
+++ b/LiveSplit/LiveSplit.Core/ComponentUtil/DeepPointer.cs
@@ -10,7 +10,7 @@ using System.Text;
 
 namespace LiveSplit.ComponentUtil
 {
-    using OffsetT = Int32;
+    using OffsetT = Int64;
 
     public class DeepPointer
     {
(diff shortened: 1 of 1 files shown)


wooferzfg:
Seems like you made your PR in your own fork rather than the LiveSplit organization's repo. I'll just merge your changes manually.

res0nance:
Doh! Sorry for the trouble

PR #1090, opened 2018-06-01

TheFuncannon:


wooferzfg:
Please change the "Type" to Script rather than Component.

Issue #1077, opened 2018-05-06
Total Playtime Oddity?


Sapphron:
I recently implemented the Total Playtime tab into my splits. However, whenever I start, the number starts counting down. It started at like 47 minutes and went down. After I waited for the timer to go all the way down, it went up to 58 minutes. Am I doing something wrong?
(sidenote: I ran the category for >8 hours before I added it)
Top image is before, bottom image is 7-8 seconds after.
![first](https://user-images.githubusercontent.com/24858043/39678713-0aea226a-5157-11e8-876a-664bee592b89.png)
![second](https://user-images.githubusercontent.com/24858043/39678714-0b0a6cdc-5157-11e8-9ba1-e3a055a2ca4f.png)

wooferzfg:
Could you send me your splits?

CryZe:
I'm pretty sure it's the 01/01/0001 bug.

wooferzfg:
For reference: https://twitter.com/wooferzfg/status/992516485292658695

Sapphron:
Where do I find the attempt history?

Sapphron:
[SMR Axeil Edition - Crocomire%.zip](https://github.com/LiveSplit/LiveSplit/files/1993029/SMR.Axeil.Edition.-.Crocomire.zip)

Sapphron:
It's not the "01/01/0001" bug.

CryZe:
It's definitely a bug with a corrupted date time in your attempt history: ![https://i.imgur.com/ngfgdT8.gif](https://i.imgur.com/ngfgdT8.gif)

hallcristobal:
it is the `01/01/0001` bug, attempt 42.

Sapphron:
woah what

Sapphron:
wait
i guess i didn't check enough

Sapphron:
it's fixed! thank you guys

CryZe:
Reopening since this seems to be a general bug that we should handle.

Sapphron:
okay, sorry

wooferzfg:
We still need to remove these broken times from the Attempt History when loading splits (similarly to how we removed negative best segment times).

@Sapphron Did you ever save an attempt midrun and choose to save it as a Personal Best?

Sapphron:
I did save attempts midrun. I haven't completed an any% run because I haven't routed it so I just skipped the rest of the times and deleted the ones that were skipped over.

[...]

PR #1098, opened 2018-06-08
Fix state being updated before reset when saving a non-PB run as a PB

wooferzfg:
Issue #1077 shows that the Attempt Ended time can sometimes be the
default value instead of the proper time. This is caused by saving a run
as a PB before the run has finished.

Still need to update the livesplit-core DLLs and update that submodule version.


wooferzfg:
Still need to update the livesplit-core DLLs and update that submodule version, but that can be done separately.

PR #1135, opened 2018-07-21
Taking ownership of the FEZ autosplitter

jbzdarkid:
@Spiraster

### DIFF
--- a/LiveSplit.AutoSplitters.xml
+++ b/LiveSplit.AutoSplitters.xml
@@ -2078,6 +2078,18 @@
   </AutoSplitter>
   <AutoSplitter>
     <Games>
+      <Game>FEZ</Game>
+      <Game>Fez</Game>
+    </Games>
+    <URLs>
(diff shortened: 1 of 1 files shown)


LiveSplit.AutoSplitters.xml
```diff
@@ -2076,6 +2076,18 @@
     <Website>https://github.com/jbzdarkid/Autosplitters</Website>
     <Description>Auto splitting and Automatic start/reset (By Darkid)</Description>
   </AutoSplitter>
+  <AutoSplitter>
+    <Games>
+      <Game>FEZ</Game>
+      <Game>Fez</Game>
```
CryZe: Game Names are case insensitive, so you have a duplicate name here.

CryZe:
Also I'd like to see some confirmation by the original author that they are okay with this ownership transfer.

Spiraster:
I'm OK with the transfer

PR #1154, opened 2018-08-12
Add Woten autosplitter

Voxelse:


Voxelse:
Why did the old commits are commited again ... Just wanted to commit the last one "Add Woten autosplitter"

PR #1185, opened 2018-10-02
Add option for static columns in Splits/Subsplits components

bgsamm:
Adds a checkbox to the column settings for both the Splits and the Subsplits components to set a column as 'static'. A static column will not be replaced by the most recent time on split, allowing for easy reference to splits from other comparisons. Only 'Split Time' and 'Segment Time' columns can be set static. Maintains backwards compatibility with saved layouts that do not have the 'Static' setting in their XML.

E.g., the 'PB' column here is set to static


bgsamm:
bump

PR #1195, opened 2018-10-18
Add Refresh Rate option to settings

ROMaster2:
The refresh rate of the timer can now be manually adjusted between 20 and 300 hertz. 20 was arbitrarily picked, but 300 was picked as it's the maximum frame rate supported by High Efficiency Video Coding. Anything higher than that will have unnoticeable real time differences.

I did not include an option for an unlocked framerate. When testing that LiveSplit locked up anytime my mouse moved over it.

### DIFF
--- a/LiveSplit/LiveSplit.Core/Options/ISettings.cs
+++ b/LiveSplit/LiveSplit.Core/Options/ISettings.cs
@@ -17,6 +17,7 @@ namespace LiveSplit.Options
 
         bool WarnOnReset { get; set; }
         bool SimpleSumOfBest { get; set; }
+        int RefreshRate { get; set; }
         IRaceViewer RaceViewer { get; set; }
         IList<string> ActiveAutoSplitters { get; set; }
         IDictionary<string, bool> ComparisonGeneratorStates { get; set; }
--- a/LiveSplit/LiveSplit.Core/Options/Settings.cs
+++ b/LiveSplit/LiveSplit.Core/Options/Settings.cs
@@ -19,6 +19,7 @@ namespace LiveSplit.Options
         public bool WarnOnReset { get; set; }
         public bool AgreedToSRLRules { get; set; }
         public bool SimpleSumOfBest { get; set; }
+        public int RefreshRate { get; set; }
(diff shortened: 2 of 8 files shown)


CryZe:
I'm not sure if we want this. LiveSplit is rendering on the CPU, and we want the least amount of CPU usage. I don't think higher framerate is worth and it will definitely lead to people complaining about the CPU usage if they increase this to 60 FPS or even higher.

ROMaster2:
I did a quick benchmark of the CPU changes. The CPU used is an [Intel Core i7-8700K overclocked to 4.7GHz](https://ark.intel.com/products/126684/Intel-Core-i7-8700K-Processor-12M-Cache-up-to-4-70-GHz-) with a monitor refresh rate of 60Hz. This was my idle and running usage with a layout that had real time, segment time, and game time displaying. The bracket number is multiplied by 12 for the single-thread usage.
```
  20 = 0.16 / 0.70   (1.92 /  8.40)
 40 = 0.19 / 1.23   (2.28 / 14.76)
 60 = 0.24 / 1.81   (2.88 / 21.72)
 80 = 0.27 / 2.48   (3.24 / 29.76)
100 = 0.29 / 2.97   (3.48 / 35.64)
120 = 0.31 / 3.00   (3.72 / 36.00)
140 = 0.33 / 3.01   (3.96 / 36.12)
...
200 = 0.38 / 3.07   (4.56 / 36.84)
...
300 = 0.49 / 3.22   (5.88 / 38.64)
```
It's viable for newer processors, but not older ones. 
It's worth noting that after 100Hz, the CPU increases diminish, probably because the window doesn't actually need to be re-rendered.

A warning could be prompted if the user tries change the FPS, or keep it as a 'hidden' setting like I suggested some time ago. The big benefit of having an increase refresh rate is re-timing accuracy. At 40Hz the margin-of-error is up to 50ms, which isn't *bad*, but could be better.

fatalis:
@ROMaster2 look into windows timer resolution. that's why

Avasam:
Any update on this? It's really weird trying to retime stuff  and not seeing the timer change every frame despite recording at 60FPS.

PR #1204, opened 2018-10-31
Ending Autosplitter

Lyfeless:
Added the autosplitter for ending to the list of splitters.


wooferzfg:
Your script URL won't work because it's not a raw file link. Please change that to the raw link, which you can get by clicking on the Raw button on the GitHub page for that file.

PR #1213, opened 2018-11-19

MrBr00d:
Change of timing method changing IGT to no load


MrBr00d:
double request, ignore this one

PR #1221, opened 2018-12-02
Add Auto Splitter for Shadow of the Tomb Raider

Atorizil:


wooferzfg:
You need to use the raw file url for the ASL script, not the "blob" link.

PR #1247, opened 2019-01-18
Iconoclasts auto splitter

seanpr96:


wooferzfg:
The type is "Component" but the download URL you provided is for a script.

PR #1251, opened 2019-01-21
Add Best Main Segments comparison

CujoIHSV:
Added an automatic comparison that groups subsplits under their corresponding top-level splits and finds the fastest total time for each top-level segment.
sobm

### DIFF
--- a/LiveSplit/LiveSplit.Core/Model/Comparisons/BestMainSegmentsComparisonGenerator.cs
+++ b/LiveSplit/LiveSplit.Core/Model/Comparisons/BestMainSegmentsComparisonGenerator.cs
@@ -1,0 +1,28 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using LiveSplit.Options;
+
+namespace LiveSplit.Model.Comparisons
+{
+    public class BestMainSegmentsComparisonGenerator : IComparisonGenerator
+    {
+        public IRun Run { get; set; }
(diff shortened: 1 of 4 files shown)


LiveSplit/LiveSplit.Core/Model/Comparisons/BestMainSegmentsComparisonGenerator.cs
```diff
@@ -0,0 +1,98 @@
...
+                    var mainSegment = mains[ind];
+                    if (mainSegment.SegmentHistory[attemptIndex][method] != null)
+                    {
+                        var currMainTime = TimeSpan.Zero;
+                        var currSubsegments = new Dictionary<ISegment, TimeSpan?>();
+                        foreach (var subsegment in subsegments[mainSegment])
+                        {
+                            var subsegmentTime = subsegment.SegmentHistory[attemptIndex][method];
+                            currSubsegments.Add(subsegment, subsegmentTime);
+                            currMainTime += subsegmentTime ?? TimeSpan.Zero;
```
wooferzfg, line 70: I don't think we're currently handling the case where the segment history element doesn't exist for the subsegment for the current attempt index.

If there's an element missing entirely, then the attempt's subsequent times won't be valid.
If there's an element with a time of `null`, then we can skip the subsegment.

You can take a look at the [Best Split Times comparison](https://github.com/LiveSplit/LiveSplit/blob/master/LiveSplit/LiveSplit.Core/Model/Comparisons/BestSplitTimesComparisonGenerator.cs#L31) for an example.

PR #1260, opened 2019-01-29
Added Resident Evil 2 Remake autosplitter

cursedtoast:


wooferzfg:
The tag name should be "URLs", not "URLS"

PR #1308, opened 2019-03-18
Fix description for Into the Breach Autosplitter

R30hedron:
Shortened to prevent the desc. from being cut off in the splits edit dialog.
Added author so that the List of ASL scripts doesn't have a '?' in the author column.


R30hedron:
Apologies for a second pull request in a single day; I didn't know that the description would get cut off...

wooferzfg:
Not a problem at all!

PR #1322, opened 2019-03-28
Add an option to always show the live delta. Requested by junkyard_dave

pgrimsrud:


CryZe:
Hey, I'll try to review this later today. This possibly clashes a bit with the changes we want to make for the split columns that we already made for LiveSplit One, but it shouldn't cause any major conflicts. We can easily transition from your changes to the new column system from LiveSplit One (if we even intend to backport it).

LiveSplit/LiveSplit.Core/Model/LiveSplitStateHelper.cs
```diff
@@ -123,7 +124,8 @@ public static class LiveSplitStateHelper
                 var bestSegmentDelta = GetLiveSegmentDelta(state, state.CurrentSplitIndex, BestSegmentsComparisonGenerator.ComparisonName, method);
                 var comparisonDelta = GetLiveSegmentDelta(state, state.CurrentSplitIndex, comparison, method);
 
-                if (splitDelta && currentTime > curSplit
+                if (liveDelta == true
```
wooferzfg: Don't need to compare to `== true` because it's already a boolean.

pgrimsrud: Fixed. I guess i'm used to being explicit about stuff like that.

pgrimsrud:
I hadn't come across LiveSplit One, and I'm not really plugged in to the current state of LiveSplit development obviously. I just jumped in and sort of made this change independently. I guess let me know if there is something else I should do, or if there are other changes I should make

CryZe:
Mmh, so it seems like the Graph and Previous Segment aren't actually gaining the setting. Only Delta, Splits and Subsplits. That's fine for me. However something that I'm considering at least slightly controversial is the following:
In LiveSplit One we already have this as a so called "Column Update Trigger". For each column you can choose when its information gets updated. Either at the beginning of the segment when entering it, contextually during the segment (like when you start losing time) or at the end of the segment when you split.
Your live delta essentially correlates with the "Start" Update Trigger. We have it per column though, while you added it as a global setting for the splits component. I think we can merge this if you make it per column instead. 

@wooferzfg thoughts?

wooferzfg:
I think it would make the most sense to:
1) Add the "update trigger" setting to other components that have different behavior where there's a "live delta".
2) Backport the split column changes and the update trigger changes from livesplit-core.

However, if #2 would take too much effort, we could just port the update trigger changes to all the components without the other changes to split columns (which is basically this PR with some modifications).

Also, why can't Graph and Previous Segment get this setting too?

CryZe:
> Also, why can't Graph and Previous Segment get this setting too?
They can, it's just that the PR doesn't actually add it. It only makes them still compile.

pgrimsrud:
I guess I wasn't sure at the time if the change made sense for Graph and Previous Segment, but I think it does and it wouldn't be too hard to add. As for points 1 and 2 I would have to look through and figure out the whole update trigger thing, but I'm guessing I don't know the codebase well enough to do that myself

CryZe:
I think the most important change is to actually make it a per column setting. From there it can be backwards compatibly changed at any points in subsequent changes (by either you or someone else).

PR #1337, opened 2019-04-12

MaximumLeech:
Added autosplitter for Trine


wooferzfg:
You need to add the AutoSplitter before the </AutoSplitters> closing tag.

PR #1342, opened 2019-04-18
Added Pat & Mat autosplitter

MisaRistolainen:
Pat & Mat autosplitter


LiveSplit.AutoSplitters.xml
```diff
@@ -5710,4 +5710,14 @@
     <Description>Load and cutscene remover (By Psych0sis and Maximum Leech)</Description>
     <Website>https://discord.gg/9Y6RK3h</Website>
   </AutoSplitter>
+  <AutoSplitter>
+    <Games>
+      <Game>Pat & Mat</Game>
```
CryZe: `&` needs to be escaped in XML as `&amp;` Same below.

PR #1344, opened 2019-04-18
Pat & Mat autosplitter

MisaRistolainen:
Pat & Mat autosplitter. Fixed & % symbols. Again...


MisaRistolainen:
Can this be merged? Seems like this was forgotten

PR #1369, opened 2019-05-08
 Added a ShortComparisonName for Best Split Times

Thermospore:


CryZe:
I'm not sure I agree with calling it "Best Pace" when shortened, as this sounds like something fundamentally different.

Thermospore:
Hmm yea I can see that. Perhaps we could use an acronym and call it `BST`?

PR #1373, opened 2019-05-13
kuso autosplitter

Zet237:
NOW kuso autosplitter for real no more xml syntax error pls!!!!


Zet237:
Added autosplitter for the game kuso.

PR #1377, opened 2019-05-19
Added Micro Mages (NES) Autosplitter

homar-mo:


wooferzfg:
You need to use the raw link for your auto splitter (by clicking on the Raw button)

PR #1400, opened 2019-06-10

mentir4242:
added boy two boy genius scripts, one for any% and one for allRedNeutron%


wooferzfg:
You can't have multiple Auto Splitters for the same game.

PR #1405, opened 2019-06-22
Saints Row 2 Autosplitter

hoXyy:


wooferzfg:
You need to escape the `&`

hoXyy:
Should be done now.

PR #1411, opened 2019-06-28
Support importing splits from URL from live split one splits.io URLs

### LINKED ISSUE #1409: Open from LiveSplit One URL
It seems like someone linked a LiveSplit One URL here that opens splits from splits.io. People were not able to use Open from URL in the original LiveSplit, as this isn't the proper splits.io URL. Maybe we should support opening from LiveSplit One URLs as well.

https://www.reddit.com/r/pathofexile/comments/bxqzq4/livesplit_for_leveling_from_fury_guide/

Blackop778:
I tried to emulate LiveSplitOne's handling of the URL without depending on the link actually starting with `https://one.livesplit.org`

Closes #1409


CryZe:
Looks good to me. Thanks for writing this :)

PR #1413, opened 2019-07-01
Adding Re-Volt Autosplitter into the list

DVark09:


wooferzfg:
You need the raw URL instead of the one you provided.

PR #1426, opened 2019-07-15
Resident Evil 4 Autosplitter / Load Remover added

Wipefinger:
Resident Evil 4 Autosplitter / Load Remover added.


NoTeefy:
@Wipefinger you need to escape the & character by &amp;

PR #1427, opened 2019-07-16
Add auto splitter for No One Lives Forever 1

poacher2k:
Shoutout to [TheTedder](https://github.com/TheTedder) for finding all the necessary pointers!


TheTedder:
Yeet

Issue #1360, opened 2019-04-29
-0.00 instead of +0.00


negative-seven:
A simple suggestion: whenever the time delta is exactly 0 (which in the case of some games with autosplitters can happen very often), mark it as being ahead (-0.00) instead of behind (+0.00), because we all need some optimism in our lives.

LukeSaward1:
You don't realize that LiveSplit can count in X.XXX (e.g. 0.000), so -0.00 could be -0.001 and +0.00 could be +0.005, for example.

Thermospore:
> [Quote: @LukeSaward1] - You don't realize that LiveSplit can count in X.XXX (e.g. 0.000), so -0.00 could be -0.001 and +0.00 could be +0.005, for example.

-7 realizes this. I don't think you fully read the issue; this part specifically: 

> whenever the time delta is exactly 0 (which in the case of some games with autosplitters can happen very often)

Issue #1336, opened 2019-04-11
Enable numbers 60 or greater in times and timer


chillex0:
e.g. allow a runner to display instead of 1:23:45, 83:45.

Excel does something similar, where in m:ss format, such a time would display as 23:45, but with a [m]:ss format would display 83:45.

PR #1446, opened 2019-08-06
Create unit tests for TimeFormatters

pengowray:
**Motivation**

Soooooooooo........ I seem to write a lot of code involving TimeSpans ([eg](https://github.com/pengowray/TimeSpanParser])), and was interested in digging into LiveSplit's internals (actually to learn more about how autosplitters work), but need to start somewhere, so I thought I'd try filling an old request which looked simple enough: 

> #1336 : Enable numbers 60 or greater in times and timer

A naive implementation might be two lines of code, and perhaps an extra user setting in an enum which might show up in a drop down. 

But it turns out there's 13 different classes which handle time formatting, and they're spread over 6 modules/submodules, and have 4 main implementation classes which are used, re-used and modified in various ways by other implementations. It's easy to see the code has grown organically over time, with a lot of duplication both between classes and within them. 

It would, in theory, be possible to go through all 13 subclasses of `ITimeFormatter` and update them to allow #1336 (At least where it's applicable. Some clearly wouldn't make sense, e.g. `PreciseDeltaFormatter` is used by the telnet server and its output format should probably never be changed). But giving more options to be used by the other formatters (those shown to the user) is currently very difficult.

My solution (which is for a later PR) is to make a single `GeneralTimeFormatter`, which can format time spans in all the ways that all the current formatters do. The `GeneralTimeFormatter` will have a few more fields, e.g. whether to show null values as `0:00` or a dash; how to display days; whether to show a plus (+) on positive times; etc. It's not that much more code than one of the larger TimeFormatters currently, and would mean all 13 of the others can be defined simply by encapsulating some parameters, rather than with a bunch of if-than-else statements. Changes to time formatting can then be made in one place, and be tested for everything at once.

I'm getting ahead of myself. This PR is just for those unit tests. These allow future changes more easily without accidentally changing anything unintentionally. Then, in future, we can look at #1336, as well as #1360, and other possibilities, like perhaps letting a speedrunner choose a different time format for in-game times (e.g. "12m13s"); or using Japanese/Chinese notation (I think something like 10分23秒), or to have a separate timer which displays frames instead of split-seconds e.g.「00:00:00;00」(which might be more meaningful than split-seconds for many games). Currently it would be a major undertaking to find and update all the appropriate TimeFormatters, but with unit tests in place, it starts to becomes much more straight forward to start considering some suggestions.

I've put the tests in groups that I found logical. Others might want to add more tests to make sure every logical branch is tested, or find a better way to arrange them or whatever.

**Tests**

* `RegularTimeFormattersTests.cs`: RegularTimeFormatter, RegularSplitTimeFormatter, RunPredictionFormatter, RegularSumOfBestTimeFormatter, AutomaticPrecisionTimeFormatter
* `ShortTimeFormatterTests.cs`: ShortTimeFormatter (with and without TimeFormat parameter), PossibleTimeSaveFormatter, SegmentTimesFormatter
* `DeltaTimeFormattersTests.cs`: DeltaTimeFormatter, DeltaComponentFormatter, DeltaSplitTimeFormatter, PreciseDeltaFormatter
* `DaysTimeFormatterTests.cs`: DaysTimeFormatter

Note: I couldn't add these to the existing `LiveSplit.Tests` subproject (I did try though), e.g. the `DataRow` attribute doesn't work with that test framework. I'm not 100% sure what the difference is between the setups of `LiveSplit.Tests` and `LiveSplit.UnitTests`, but anyway, they do seem to be logically different enough to warrant separate projects. The existing test is testing the contents of the AutoSplitters xml file and not testing code, so it's a test but not strictly a unit test. 

**Changes to other modules**

Most of the TimeFormatter classes tested are `internal` (i.e. not `public` nor `private`). To make unit tests for them with minimal changes, this line has to be added to `AssemblyInfo.cs` :
```C#
[assembly: InternalsVisibleTo("LiveSplit.UnitTests")]
```
...of these submodules:
* LiveSplit.Delta
* LiveSplit.RunPrediction
* LiveSplit.Server
* LiveSplit.Splits
* LiveSplit.SumOfBest

~~I guess I'll make~~ I've now made a bunch of PR's for those, as it's a required change for this PR to work.

Note, these ones already use public classes so require no changes:
* LiveSplit.TotalPlaytime (DaysTimeFormatter)
* LiveSplit.DetailedTimer (SegmentTimesFormatter)


wooferzfg:
These tests should really be in `LiveSplit.Tests`. We have so few tests already so there's no good reason to have two separate projects for them. I'm not sure why `DataRow` wouldn't work, but I don't think that justifies making another project.

Also, the Time Formatter classes should all be `public`. I would suggest making that change rather than exposing the internals.

pengowray:
OK I've merged UnitTests and Tests, and made the formatter classes public. I've also renamed some formatters because they have identical names + namespaces in two projects. (splits & subsplits).  It's quite difficult working with so many submodules in their own repos.

wooferzfg:
Yeah, it's frustrating and I think it was a bad design choice to have all the components be submodules. We only really should've done that for components that are forked from other repos.

LiveSplit/LiveSplit.sln
```diff
@@ -113,6 +113,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "LiveSplit.Counter", "Compon
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "LiveSplit.BlankSpace", "Components\LiveSplit.BlankSpace\LiveSplit.BlankSpace.csproj", "{825B6521-94B8-4C8A-9709-3CA2ABFE70D7}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "LiveSplit.UnitTests", "UnitTests\LiveSplit.UnitTests.csproj", "{80866AD2-578B-439A-B744-62C05F13005A}"
+EndProject
```
wooferzfg: Can you remove `LiveSplit.UnitTests` from `LiveSplit.sln`?

wooferzfg:
Also, before this is merged, I would like to see a successful build that includes the commits from the different submodules. You should be able to update `.gitmodules` to target a commit on your fork. Once we get a passing build, we can undo the changes to `.gitmodules` and then merge the pull request for each component.

PR #1454, opened 2019-08-11
Update Call of Duty 4: Modern Warfare autoSplitter

KunoDemetries:
Changing out jukspa's load remover and adding an auto splitter to it; it also includes the exit split.

I forgot to remove his line of code from the list but for the OG one it's on the line number 2631; on mine, it's about 2642.

### DIFF
--- a/LiveSplit.AutoSplitters.xml
+++ b/LiveSplit.AutoSplitters.xml
@@ -43,6 +43,17 @@
 	<Type>Script</Type>
 	<Description>Basic autosplitter and load remover. (By Jayrude)</Description>
 	<Website>https://discord.gg/wUYXU4J</Website>
+		<AutoSplitters>
+<AutoSplitter>
+	<Games>
+		<Game>Call of Duty 4: Modern Warfare</Game>
+	</Games>
+	<URLs>
+		<URL>https://raw.githubusercontent.com/KunoDemetries/cod4/master/mw.asl</URL>
+	</URLs>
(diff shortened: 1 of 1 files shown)


LiveSplit.AutoSplitters.xml
```diff
@@ -43,6 +43,17 @@
 	<Type>Script</Type>
 	<Description>Basic autosplitter and load remover. (By Jayrude)</Description>
 	<Website>https://discord.gg/wUYXU4J</Website>
+		<AutoSplitters>
```
wooferzfg: This tag shouldn't be here.

LiveSplit.AutoSplitters.xml
```diff
@@ -43,6 +43,17 @@
...
+<AutoSplitter>
+	<Games>
+		<Game>Call of Duty 4: Modern Warfare</Game>
+	</Games>
+	<URLs>
+		<URL>https://raw.githubusercontent.com/KunoDemetries/cod4/master/mw.asl</URL>
+	</URLs>
+	<Type>Script</Type>
+	<Description> Autosplitter and load remover for Call of Duty 4: Modern Warfare on Steam. Does a exit split after crosshair dissapears. (by KunoDemetries)</Description>
+	<Website>https://discord.gg/fQaYqvFF</Website>
```
wooferzfg: You need a `</AutoSplitter>` tag after this line.

PR #1457, opened 2019-08-13
Simplify and centralize time formatters

pengowray:
This creates a "GeneralTimeFormatter" class, which is capable of formatting a TimeSpan in all the ways all the existing ITimeFormatter classes do. This reduces duplicate code, keeps things consistent more easily, makes future changes to time formatting possible, and reduces minor bugs (e.g. negative TimeSpans are always handled correctly now without dropping the hours, which a couple of formats did previously)

This is purely a refactoring. There are no user-facing changes (or API changes).

This PR also updates all the Time Formatters that are in this repo to inherit from GeneralTimeFormatter and reduces their code to a few lines each. The unit tests all still pass. 

The time formatters in the submodules can get similar treatment but I haven't committed those yet. Those ones still work with these changes, even when they're calling formatters in the main repo (because the Format() methods of all refactored classes still give the same output).

Some notable code changes:

* added "SingleMinutes" to the TimeFormat enum. This is for time formats like "0:12" (for 0m12s), which is used by the RegularTimeFormatter classes. I've added it to the enum but it remains hidden in the UI so users can't select it for components that didn't have it before.

* created a [NullFormat enum](https://github.com/LiveSplit/LiveSplit/commit/cd4f4771c8365ccb535e89227a0f958dc2994db8#diff-9cce6fc4df1fb15b5f1ad90a37cc370b), to choose how null `TimeSpan?` values are displayed (typically a dash or some sort of zero value). I've kept all the Time Formatters returning exactly what they did before. Someone might want to change them about in future for nicer/more appropriate display (e.g. might want to display null times from speedrun.com as `-:--` instead of as `0.00`; and many probably never format null values regardless) But in this PR the output is the same.

### DIFF
--- a/LiveSplit/LiveSplit.Core/TimeFormatters/PossibleTimeSaveFormatter.cs
+++ b/LiveSplit/LiveSplit.Core/TimeFormatters/PossibleTimeSaveFormatter.cs
@@ -2,27 +2,12 @@
 
 namespace LiveSplit.TimeFormatters
 {
-    public class PossibleTimeSaveFormatter : ITimeFormatter
+    public class PossibleTimeSaveFormatter : GeneralTimeFormatter
     {
-        public TimeAccuracy Accuracy { get; set; }
-
-        public string Format(TimeSpan? time)
+        public PossibleTimeSaveFormatter() : base()
         {
-            var formatter = new ShortTimeFormatter();
-            if (time == null)
-                return TimeFormatConstants.DASH;
-            else
-            {
-                var timeString = formatter.Format(time);
-                if (Accuracy == TimeAccuracy.Hundredths)
-                    return timeString;
-                else if (Accuracy == TimeAccuracy.Tenths)
-                    return timeString.Substring(0, timeString.Length - 1);
-                else
-                    return timeString.Substring(0, timeString.Length - 3);
-
-            }
-                
+            Accuracy = TimeAccuracy.Hundredths;
+            NullFormat = NullFormat.Dash;
         }
     }
 }
--- a/LiveSplit/LiveSplit.Core/TimeFormatters/PossibleTimeSaveFormatter.cs
+++ b/LiveSplit/LiveSplit.Core/TimeFormatters/PossibleTimeSaveFormatter.cs
@@ -4,7 +4,7 @@ namespace LiveSplit.TimeFormatters
 {
     public class PossibleTimeSaveFormatter : GeneralTimeFormatter
     {
-        public PossibleTimeSaveFormatter() : base()
+        public PossibleTimeSaveFormatter()
         {
             Accuracy = TimeAccuracy.Seconds;
             NullFormat = NullFormat.Dash;
--- a/LiveSplit/LiveSplit.Core/TimeFormatters/PossibleTimeSaveFormatter.cs
+++ b/LiveSplit/LiveSplit.Core/TimeFormatters/PossibleTimeSaveFormatter.cs
@@ -6,7 +6,7 @@ namespace LiveSplit.TimeFormatters
     {
         public PossibleTimeSaveFormatter() : base()
         {
-            Accuracy = TimeAccuracy.Hundredths;
+            Accuracy = TimeAccuracy.Seconds;
             NullFormat = NullFormat.Dash;
         }
     }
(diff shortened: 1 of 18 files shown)


LiveSplit/LiveSplit.Core/TimeFormatters/TimeFormat.cs
```diff
@@ -2,6 +2,6 @@
 {
     public enum TimeFormat
     {
-        TenHours, Hours, Minutes, Seconds
+        TenHours, Hours, Minutes, Seconds, SingleMinutes
```
CryZe: I'd like to unify the Time Formatting as much as possible across the original LiveSplit and LiveSplit One / livesplit-core. So I'll likely apply the changes in this branch to that codebase in some form. One thing I'd somewhat like backported, which you already slightly started here, is all the other variants as well: https://github.com/LiveSplit/livesplit-core/blob/master/src/timing/formatter/digits_format.rs

CryZe:
Similarly, we probably want the Milliseconds Accuracy here as well: https://github.com/LiveSplit/livesplit-core/blob/master/src/timing/formatter/accuracy.rs

LiveSplit/LiveSplit.Core/TimeFormatters/TimeFormat.cs
pengowray: I've removed "SingleMinutes" again and added a whole new `DigitsFormat` enum which is the same core's, which GeneralTimeFormatter will now use. I didn't update or replace TimeFormat as there are too many components / settings files / etc that would need to be updated or might potentially break. So TimeFormat is still an option for formatting, but it's marked as deprecated, and using it with GeneralTimeFormatter automatically converts over to DigitsFormat now.

Also added Milliseconds accuracy (0.000). Also a bunch of unit tests for formatting with this new TimeAccuracy, as well as all the new `DigitsFormat` values.

Also discovered I hadn't tested "AutomaticPrecision" and my implementation was broken, so copied back some original code for handling that.

Note you can choose between hours or days as the largest unit to include, but still not minutes (#1336).

LiveSplit/LiveSplit.Core/TimeFormatters/PossibleTimeSaveFormatter.cs
```diff
@@ -2,27 +2,12 @@
 
 namespace LiveSplit.TimeFormatters
 {
-    public class PossibleTimeSaveFormatter : ITimeFormatter
+    public class PossibleTimeSaveFormatter : GeneralTimeFormatter
     {
-        public TimeAccuracy Accuracy { get; set; }
-
-        public string Format(TimeSpan? time)
+        public PossibleTimeSaveFormatter() : base()
```
wooferzfg: Why do you have `: base()` on some but not all of these classes? Is it actually necessary?

LiveSplit/LiveSplit.Tests/TimeFormatTests/TimeAccuracyTests.cs
```diff
@@ -0,0 +1,42 @@
...
+        [DataRow("00:00:10.8888888", "0:10.88", TimeAccuracy.Hundredths)]
+        [DataRow("00:00:11.8888888", "0:11.888", TimeAccuracy.Milliseconds)]
+
+        public void TestTimeAccuracy(string timespanText, string expected, TimeAccuracy accuracy)
+        {
+            var formatter = new GeneralTimeFormatter();
+            formatter.DigitsFormat = DigitsFormat.SingleDigitMinutes;
+            formatter.Accuracy = accuracy;
+
+
```
wooferzfg: You have a lot of extra newlines in your code. Could you get rid of them?

wooferzfg:
@CryZe I think it would be cool if we copied these unit tests over to `livesplit-core` and then did the same type of refactoring to one generic time formatter.

LiveSplit/LiveSplit.Core/TimeFormatters/PossibleTimeSaveFormatter.cs
pengowray: Not necessary. Can remove. Left over because I was originally going to give GeneralTimeFormatter a bunch of other constructors.

LiveSplit/LiveSplit.Tests/TimeFormatTests/TimeAccuracyTests.cs
pengowray: Yep

PR #1460, opened 2019-08-15

makkebakke:
Adds Preflight Panic and Brütal Legend autosplitters.


makkebakke:
Please double-check symbol definition of "ü".

PR #1467, opened 2019-08-17
Updating autostart and loadless for DyingLight

shadow2hel:


wooferzfg:
You need to replace the `&` with `&amp;`

PR #1488, opened 2019-09-16
Fix blocking icons files on change

Moliman:
This fix should prevent blocking file when we use it as an icon, if LiveSplit is not closed


wooferzfg:
I had to revert this because this solution breaks animated GIFs.

PR #1496, opened 2019-09-26
Add EDF5 autosplitter

Souzooka:


wooferzfg:
You need to add your autosplitter before the `</AutoSplitters>` tag.

Souzooka:
Overlooked that, should be fixed now.

PR #1521, opened 2019-10-16
Replace Serious Sam TFE/TSE Autosplitters

apple1417:
Tezur0 ok'd this, as he should confirm below


Tezur0:
Confirming that I'm okay with apple replacing my asl's with his.

PR #1543, opened 2019-11-10

CallumMcW:
Updated the Load Removal for QUBE 2, as the old one no longer worked on the recent update and the community member who made it is now inactive in the community

Edit: Also added some debug output stuff for the qube2 load remover


wooferzfg:
That download link won't work. You need to use a raw download URL.

CallumMcW:
Sorry, wasn't thinking when I pasted it in, fixed now

PR #1564, opened 2019-12-14
doc/Auto-Splitters: Better detail when isLoading affects the timer

SirGFM:


CryZe:
I'm not sure if it makes sense to document it there as this isn't user facing but developer facing documentation. But it certainly doesn't hurt to have it here, so I'll merge it. Thanks :)

PR #1586, opened 2020-01-10
Add GTA V autosplitter

hoXyy:


wooferzfg:
There's already a GTA 5 autosplitter. If you want to replace it, you'll need to remove the existing entry from LiveSplit.AutoSplitters.xml.

PR #1591, opened 2020-01-14
Timermodel updatetimes method

TheTedder:
@CryZe 
Added a public UpdateTimes() method that wraps the functionality of the four UpdateXxxx() methods already present. I did this because my multiruns plugin was breaking without access to these methods which were privated in the 1.7.7 update.


wooferzfg:
@TheTedder Do you need a 1.7.8 update with these changes, or are you ok with just using the dev build?

TheTedder:
@wooferzfg I would like to have a 1.7.8 if possible so I can recompile my plugin for people to use with release LiveSplit. It's not a huge deal though. If there's other outstanding stuff that you would like to have done before putting out another version let me know and I'll work on it.

PR #1605, opened 2020-01-21
Updated FEAR asl

KunoDemetries:
Changed a lot of loading rules, and updated the asl to split faster


wooferzfg:
The type should be `Script` instead of `Component`.

PR #1613, opened 2020-01-23
Added Autosplitter for F-Zero

remigranotier:
Compatible with Ultime Decathlon 8's pack's version of snes9x.
Will probably not work with bizhawk.

### DIFF
--- a/LiveSplit.AutoSplitters.xml
+++ b/LiveSplit.AutoSplitters.xml
@@ -7374,3 +7374,15 @@
     <Website>https://github.com/Voxelse/ASLScripts/tree/master/Livesplit.BubbleGhost</Website>
   </AutoSplitter>
 </AutoSplitters>
+  <AutoSplitter>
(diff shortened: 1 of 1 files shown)


LiveSplit.AutoSplitters.xml
```diff
@@ -7373,4 +7373,16 @@
     <Description>Auto Splitting and Auto Start/Reset are available (By Voxelse)</Description>
     <Website>https://github.com/Voxelse/ASLScripts/tree/master/Livesplit.BubbleGhost</Website>
   </AutoSplitter>
+</AutoSplitters>
+  <AutoSplitter>
+    <Games>
+      <Game>F-Zero</Game>
+    </Games>
+    <URLs>
+      <URL>https://github.com/remigranotier/Autosplitters/blob/master/FZero.asl</URL>
```
wooferzfg: This needs to be a raw link instead.

PR #1614, opened 2020-01-24
Autosplitter for F-Zero - Raw content now (sorry)

remigranotier:
Cannot reset on first lap of race of between races


LiveSplit.AutoSplitters.xml
```diff
@@ -7374,3 +7374,15 @@
     <Website>https://github.com/Voxelse/ASLScripts/tree/master/Livesplit.BubbleGhost</Website>
   </AutoSplitter>
 </AutoSplitters>
+ <AutoSplitter>
```
wooferzfg: You added this autosplitter in the wrong spot. It should be above the closing `</AutoSplitters>` tag.

PR #1664, opened 2020-02-27
Tomb Raider: Anniversary autosplitter

NextLevelMemes:
Autosplitter with 23 autosplits, autoreset and load removal.


CryZe:
You can just keep editing the file, you don't need to reopen the PR all the time.

PR #1690, opened 2020-03-15
Click through window

esseivan:
I had some problems in some games where the mouse is not visible, but sometimes get in the window of LiveSplit.
Then if I would left click, the game would not be in focus anymore :/

So I made a fix for issues like this :
I added a setting "Click through window on run"
Which means that if enabled, when the timer is running, clicking on the LiveSplit window will click through it.

The changes I made :
I added a checkbox in the settings window
I added a event listener for TimerModel for almost all events and check if the currentPhase is running or not. Whether it is running (and the setting is enabled), it calls the function SetClickThrough from TimerForm

The code for the last function (SetClickThrough) is inspired from https://stackoverflow.com/questions/2798245/click-through-in-c-sharp-form

You may notice some strange edits in SettingsDialog.Designer.cs but this is Visual Studio doing some refactoring when editing the design of the form

Edit . The game I spoke about was Golf It!  - The mouse is invisible but not stuck in the middle of the screen, so it would eventually get in the LiveSplit's window

### DIFF
--- a/LiveSplit/LiveSplit.Core/Options/HotkeyProfile.cs
+++ b/LiveSplit/LiveSplit.Core/Options/HotkeyProfile.cs
@@ -20,6 +20,7 @@ namespace LiveSplit.Options
         public bool GlobalHotkeysEnabled { get; set; }
         public bool DeactivateHotkeysForOtherPrograms { get; set; }
         public bool DoubleTapPrevention { get; set; }
+        public bool ClickThrough { get; set; }
 
         public const string DefaultHotkeyProfileName = "Default";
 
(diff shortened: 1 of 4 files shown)


wooferzfg:
How is this different from https://github.com/LiveSplit/LiveSplit/pull/882?

esseivan:
Ah I didn't noticed this was already a thing.
I used v1.7.7 and there's no such checkbox in layout settings, not sure why tho
Here's what I have
image

But then you're right, they're the same

wooferzfg:
That feature is not in 1.7.7 because we have only included bug fixes in recent releases. It is available in the [dev build](https://raw.githubusercontent.com/LiveSplit/LiveSplit.github.io/artifacts/LiveSplitDevBuild.zip), however.

esseivan:
Oh ok thanks
I think we can close this then

esseivan:
Oh and btw, I had an exception at IRun.cs line 77 : Sequence contains no elements
(More precisely IRun.cs line 100 because SegmentHistory is empty)
I already had it while debugging via VS
and with the dev build you just linked me, I still have this exception.
I don't know if it's already a known exception or not

PR #1701, opened 2020-03-23
Rockman/Mega Man ZX Advent

Ikkisoad:
Auto splitters for Rockman and Mega Man ZX Advent on DeSmuME and Zero/ZX Legacy Collection (PC).


wooferzfg:
You need a raw link to the file.

PR #1707, opened 2020-03-26
Use splitter for catext/other associated games

jbzdarkid:


LiveSplit.AutoSplitters.xml
```diff
@@ -2570,6 +2570,8 @@
       <Game>Road to Gehenna</Game>
       <Game>Gehenna</Game>
       <Game>The Talos Principle Category Extensions</Game>
+      <Game>The Talos Principle Custom Maps</Game>
+      <Game>The Talos Principle VR</game>
```
wooferzfg: This needs to be `</Game>`

Issue #952, opened 2017-12-15
Add "Resume Run" feature.


Kigoli:
This feature will save the paused time and use a splits file to resume the run using the completed splits as well as the paused time.

Kigoli:
#953

MinecraftPlaye:
@Kigoli any news on this one?

Kigoli:
@MinecraftPlaye

I got it working, but it was definitely a hack job.

Because it wasn't up to the projects standards, it wasn't committed to the office project. 

This was a side project for me while I had some vacation time to burn.

If I ever feel inclined, I might pick it back up.

Kigoli:
@MinecraftPlaye

I got it working, but it was definitely a hack job.

Because it wasn't up to the projects standards, it wasn't committed to the office project. 

This was a side project for me while I had some vacation time to burn.

If I ever feel inclined, I might pick it back up.

gth:
Perhaps to get the feature implemented at all, remove the requirement for hibernate to support autosplitters?  i.e., the hibernate button would be disabled if an Autosplitter is detected.

CryZe:
Yeah I believe for an initial implementation the auto splitters should be out of scope. So I guess I‘d be open to some prototypical implementation and we‘ll figure it out from there. Keep in mind that it probably will have to change a bit before we find a solution we are happy with, so be prepared for changes that are possibly required then.

spenrose:
I am also trying to use LiveSplit to track progress over several days while leveling in Path of Exile. I see you have some code pushed now and would be happy to test it if you need some feedback/testing.

DarkRTA:
We should also have an option to undo the pause in case this is somehow used as a preventive measure against LiveSplit crashing.

In this scenario we'd back up the currently active run after every split and if LiveSplit crashes for some reason, offer to restore that run upon the next launch.

Granted in most cases its possible to still finish a run even after the timer crashes.

shane1595042264:
guys I really want this feature

ThatGuyMike7:
same, need this for tracking time in casual Path of Exile runs

Brambor:
My idea for this feature: Just save on closing. When launching, if you don't want the last run, press reset.

It should be default baked in like that.

Shotnex4:
Hey everyone, I also missed this feature and tried to solve it myself.
[...]

PR #1774, opened 2020-04-20
Issue #952 - Added hibernate and resume run

goodtrailer:
#952 

Adds:
- Hibernating (saving) a paused run to a .lsr (LiveSplit Run) file that is formatted in XML
- Loading from a .lsr as long as the timer isn't running. After loading the run, the timer is still paused, it doesn't immediately resume.
- Appropriate right click menu buttons (under "Controls") and hotkey settings (unbound by default)

Notes:
- Disabled if autosplitter is active
- Doesn't change any existing methods
- Exception handling might be a little spotty
- Possible confusion in method/variable naming, because while the word "run" is usually used to mean a specific speedrun, in the code the `Run` class is used to store all of the past runs, metadata, game name, category, etc. For example, the new `TimerModel.SaveRun()` isn't actually used to save all the data in the `Run` class.


LiveSplit/LiveSplit.Core/Model/TimerModel.cs
```diff
@@ -144,6 +145,53 @@ private void ResetSplits()
             CurrentState.Run.FixSplits();
         }
 
+        public void LoadRun(string gameName, string categoryName, Time time, Dictionary<string, Time> segments, AtomicDateTime started, bool isGameTimeInitialized, TimeSpan pauseTime)
+        {/*
+            string msg = "gameName: " + gameName + "\ncategoryName: " + categoryName + "\ntime: " + time.ToString() + "\nsegments: ";
```
kkaja123, line 150: Dead code should be removed unless it has a reason to stay for informative reasons. If so, provide a comment stating why it has been disabled but kept.

kkaja123:
Maintainers, this feature has been requested several times in issues, so why has it been ignored/overlooked?

LiveSplit/LiveSplit.Core/Model/ITimerModel.cs
```diff
@@ -27,6 +28,7 @@ public interface ITimerModel
         void UndoSplit();
         void Reset();
         void Reset(bool updateSplits);
+        void LoadRun(string gameName, string categoryName, Time time, Dictionary<string, Time> segments, AtomicDateTime started, bool isGameTimeInitialized, TimeSpan pauseTime);
```
wooferzfg, line 31: This is too many parameters for a method. This should take some kind of class or struct instead.

LiveSplit/LiveSplit.View/View/TimerForm.cs
```diff
@@ -2556,6 +2579,164 @@ private void Reset()
...
+            catch (FormatException)
+            {
+                MessageBox.Show("The Livesplit Run (.lsr) file is invalid. One or more values is formatted improperly.", "Livesplit Run File Error", MessageBoxButtons.OK, MessageBoxIcon.Error);
+            }
+            catch (MismatchedGameCategoryException)
+            {
+                MessageBox.Show("The Livesplit Run (.lsr) file is for a different game or run category.", "Livesplit Run File Error", MessageBoxButtons.OK, MessageBoxIcon.Error);
+            }
+            catch (IndexOutOfRangeException)
+            {
```
wooferzfg, line 2735: "Livesplit" -> "LiveSplit"

LiveSplit/LiveSplit.View/View/TimerForm.cs
```diff
@@ -2556,6 +2579,164 @@ private void Reset()
             }
         }
 
+        private void HibernateRun()
```
wooferzfg, line 2582: This should not all live in `TimerForm`.

wooferzfg (commented):
I'm not sure that having a separate file for an in progress run is a good idea. It's a bit unintuitive to have to load a splits file and *then* load a hibernated run. Including this data in the splits file might be better.

On top of that, it might be more convenient if "Save Splits" prompted you about hibernating the in progress run if you saved in the middle of the run.

In general, we also need to make sure that we have a design that works with livesplit-core before we decide to merge this.

CryZe:
Yeah I'd prefer it as well if we don't have a new file format for this.

goodtrailer:
I agree that in-progress runs should be hibernated into the splits file instead of a separate .lsr, but I'm not sure that's possible without making changes to livesplit-core, since `RunFactory` relies on the extern `Run_parse`. And since modifying livesplit-core is probably the proper way to go about this, most of what I've written here seems to be the wrong approach. (Maybe I'll see if I can make those changes to livesplit-core, but I'm pretty unfamiliar with Rust.)

benshemmeld:
@goodtrailer I'm really eager to get this feature merged into the main project. I've been using your code for a few days (and fixed one issue related the date/time format, because I'm in Australia). I was just wondering if you're planning on making the suggested changes to this pull request? If not, I'd be happy to make the changes myself, and put up a new pull request.

goodtrailer:
@benshemmeld I was planning on looking into livesplit-core when I found the time, but if you do find a way to make the code save to the splits file or want to implement the other suggested changes feel free to do so, since I'm fairly busy for the time being.

daanbom:
Hey, I had a question. Is there a way to use this as i could really use this in my streams the few upcoming weeks

shane1595042264:
I am just wondering if it's possible to just automate the process of setting "the timer start at" the current time? Because that's how I hibernate my run manually. It would very convenient if it's all in one button so that I don't have to type all the numbers again.

daanbom:
> I am just wondering if it's possible to just automate the process of setting "the timer start at" the current time? Because that's how I hibernate my run manually. It would be very convenient if it's all in one button so that I don't have to type all the numbers again.

yeah also this doesnt work with splits

shane1595042264:
What a pity.

PR #1776, opened 2020-04-20
Adding Zed Autosplitter

DanfsAC:


DanfsAC:
Well, I have no idea what exactly went wrong. If someone can at least explain the problem, I would appreciate.

wooferzfg:
There was an issue with the CI. It should be resolved now.

PR #1812, opened 2020-05-05
Share to Twitch Fix

Permamiss:
This fixes the "Share to Twitch" feature of LiveSplit.

- Add new file to LiveSplit.Register
  - Code to easily add "LiveSplit.exe" to Registry, so Twitch will load properly in emulated browser
- Update Program.cs to utilize said file to add DWORD "LiveSplit.exe" to Registry if not already there
- Update OAuth code in Twitch.cs to match Kraken/v5 requirements
  - Change from ChannelName to ChannelId
  - Game Search change from "q" to "query"; remove deprecated "&type=suggest"

Thanks to @hlepouse for their help!


LiveSplit/LiveSplit.Register/InternetExplorerBrowserEmulation.cs
```diff
@@ -0,0 +1,272 @@
...
+        /// <summary>
+        /// Interner Explorer 11 Edge Mode
+        /// </summary>
+        Version11Edge = 11001
+    }
+
+    /// <summary>
+    /// Helper methods for working with Internet Explorer browser emulation
+    /// </summary>
+    internal static class InternetExplorerBrowserEmulation
```
wooferzfg, line 76: I'm confused why we need this. We're able to open a browser window for [Speedrun.com OAuth](https://github.com/LiveSplit/LiveSplit/blob/master/LiveSplit/LiveSplit.View/View/SpeedrunComOAuthForm.cs). Is this different enough that we can't do that? Either way, this solution is ugly, and I think we should figure out something that doesn't require emulating IE.

Permamiss: Anyone can feel free to do that! This is simply patching the existing feature where LiveSplit utilized an emulated browser. Twitch does not work with Compatibility View, which emulated browsers have by default unless you add a Registry key. Multiple apps utilize this function, including Zoom, RuneScape, TeamViewer, etc.

hlepouse: "We're able to open a browser window for Speedrun.com OAuth"

I don't understand what you're talking about, can you calrify please ? From what I see, LiveSplit authentification to speedrun.com is done through System.Windows.Forms.WebBrowser as well

wooferzfg: Sorry, I got confused. I see what's happening - we added a class with helpers that add a registry key because Twitch doesn't work in compatibility view for browser emulation, while Speedrun.com does.

My question is - how hard would it be to get Twitch OAuth working without needing to modify the registry? I realize that you're just making the existing code work, but we might want to reconsider using an emulated browser if this is what it requires.

LiveSplit/LiveSplit.Register/InternetExplorerBrowserEmulation.cs
```diff
@@ -0,0 +1,272 @@
...
+    // Configuring the emulation mode of an Internet Explorer WebBrowser control
+    // http://cyotek.com/blog/configuring-the-emulation-mode-of-an-internet-explorer-webbrowser-control
+
+    // Additional Resources:
+    // http://msdn.microsoft.com/en-us/library/ee330730%28v=vs.85%29.aspx#browser_emulation
+
+    /// <summary>
+    /// Internet Explorer browser emulation versions
+    /// </summary>
+    public enum BrowserEmulationVersion
```
wooferzfg, line 20: Could you make any enums or methods in this class private unless we actually call them from outside of this class?

Permamiss: Ah, good call. I'm changing the following:
**private**
- GetBrowserEmulationVersion()
- GetInternetExplorerMajorVersion()
**internal**
enum BrowserEmulationVersion
IsBrowserEmulationSet()
SetBrowserEmulationVersion() (both of them)

LiveSplit/LiveSplit.Register/InternetExplorerBrowserEmulation.cs
Permamiss: If you're talking about getting Twitch OAuth to work while still utilizing an emulated browser, it's not possible from what I have seen. This is the least "hacky" solution out there; one of them recommended grabbing the page's HTML contents and inserting a header to pretend the browser is a different version.
I am all for looking into opening the user's default browser, doing the OAuth through there, and somehow passing it back into LiveSplit, but before that I think we should start with fixing what was broken.

Permamiss:
Is there anything else that needs to be changed before merging this with the master branch?

wooferzfg:
Nope, this looks good. I just need to test it, which I'll try to do soon.

Permamiss:
Great! Could someone **Approve** the [requested changes](https://github.com/LiveSplit/LiveSplit/pull/1812/files/f7fcb558d7b9c592f8f6d80feb0586ea2abf0bb4) since it has been resolved and the author (me) is not allowed to approve the changes? :)

PR #1856, opened 2020-05-23
Add start offset parameter to core timer method

deains:
This is a small update to the core API to allow for a more elegant fix for the racetime.gg integration with the main timer.

This update allows the timer to be started with a supplied offset value, which can be either positive or negative. If not supplied, the behaviour of this method is unchanged (same as suppling an offset value of zero).

### DIFF
--- a/LiveSplit/LiveSplit.Core/Model/DoubleTapPrevention.cs
+++ b/LiveSplit/LiveSplit.Core/Model/DoubleTapPrevention.cs
@@ -65,11 +65,11 @@ namespace LiveSplit.Model
             return TimeStamp.Now - LastEvent > Delay;
         }
 
-        public void Start()
+        public void Start(TimeSpan? start = null)
         {
             if (CheckDoubleTap())
             {
-                InternalModel.Start();
+                InternalModel.Start(start);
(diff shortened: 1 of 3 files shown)


wooferzfg:
The Racetime integration shouldn't be overriding the timer offset at all. That value is important for some games where the timer doesn't start at 0, and it also causes problems for Autosplitters. The only automatic actions that the Racetime integration should be doing is splitting, unsplitting, or resetting.

deains:
@wooferzfg I've tested this with an autosplitter and it didn't interfere for me. I've also tested it with setting the timer to not start at 0, as you can see in the PR that is included in the timer calculation so I don't believe there'd be any issue around that.

I've tested as much as I can with this and can't find any problems but it's your call. Let me know if you still want to nix this feature.

wooferzfg:
This won't work if there's an Autosplitter for a game where the timer runs at 3x speed, for example. Also, this PR would change the interface for TimerModel, which would break a lot of existing components that haven't been recompiled in years.

deains:
Alrighty, fair enough. I'll update the racetime PR. 🙂

PR #1865, opened 2020-05-27
Years and months

mellester:
image

I updated the timer and detailed timer to handle longer timespans


kikorp78 (commented):
Years, days? In my opinion, that is not really needed as this is used for speedruns, not for something that could take years, or even days.

mellester:
Some in game counters are coutned in years and days

PR #1873, opened 2020-06-04
Changing Desc of Wizard of Legend

QunurZ:
Forgot to change the description of Wizard of Legend.


QunurZ:
Don't know why it failed since i only changed the desc of the Wizard of Legend

wooferzfg:
You can't use `&`. You need to use `&amp;`.

QunurZ:
Thanks, had my suspicion but wasn't sure.

PR #1880, opened 2020-06-10
Update LiveSplit.AutoSplitters.xml for Ys 1

Meta4ever:
adding Ys 1


LiveSplit.AutoSplitters.xml
```diff
@@ -1,5 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <AutoSplitters>
+	<Autosplitter>
```
wooferzfg, line 3: This needs to be `<AutoSplitter>`

PR #1913, opened 2020-06-30

tocaloni1:


wooferzfg:
the `AutoSplitter` tag needs to be before the `</AutoSplitters>` closing tag

PR #1982, opened 2020-08-01

Mysterion06:


Mysterion06:
Yo wooferzfg, you just verified it, but why is it not available in Livesplit?

wooferzfg:
It should be. Check the autosplitters XML file in your LiveSplit folder to see whether it includes your changes.

PR #2025, opened 2020-09-24
added crysis remastered asl, updated far cry asl path

Skiendys:
updated far cry asl path to new one and added crysis remastered script

### DIFF
--- a/LiveSplit.AutoSplitters.xml
+++ b/LiveSplit.AutoSplitters.xml
@@ -6220,10 +6220,10 @@
             <Game>FC</Game>
         </Games>
         <URLs>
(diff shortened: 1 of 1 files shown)


wooferzfg:
Oops, shouldn't have merged that.

wooferzfg:

Skiendys:
> We're recently switched repositories for the Auto Splitters XML! Please open a PR in [this repo](https://github.com/LiveSplit/LiveSplit.AutoSplitters) instead. The most recent version of LiveSplit uses the new repository, and the XML will be automatically be committed to the old repository every 24 hours for compatibility with older versions.

oh thanks for info will be updating other one from now on :)

PR #2026, opened 2020-09-24
Add feature to split layout into several rows/columns

tduva:
I added this feature for a stream layout where two rows are required, each with it's own subsplits instance, one showing the overview of the run and the other showing the individual splits. Without this, two instances of LiveSplit are required to make it work (it would be possible to have one really wide LiveSplit instance and just split up the window capture, however it is required to be on a screen to render, so it doesn't work in every situation).

This feature modifies the ComponentRenderer to start a new line (column for vertical layout, row for horizontal layout) at every SplitAtComponent. The size (width for vertical layout, height for horizontal layout) of each line is reduced accordingly. The SplitAtComponent isn't implemented as a regular component, because it doesn't really do anything on it's own (except handling settings) and is tied into a core function, so it has some special handling for adding and loading.

For the user however, the component can simply be added like a regular component in the layout editor (in the component menu under "Other"). It also has settings to change the relative size of a line and the spacing to the previous line.

Horizontal Layout:
LiveSplit_Lines_Example

Vertical Layout:
LiveSplit_Lines_Example2

Layout Editor:
LiveSplit_Lines_Layout

### DIFF
--- a/LiveSplit/LiveSplit.Core/LiveSplit.Core.csproj
+++ b/LiveSplit/LiveSplit.Core/LiveSplit.Core.csproj
@@ -85,6 +85,7 @@
     </Reference>
     <Reference Include="System" />
     <Reference Include="System.Core" />
+    <Reference Include="System.Data" />
     <Reference Include="System.Drawing" />
     <Reference Include="System.Web" />
     <Reference Include="System.Web.Extensions" />
@@ -179,6 +180,14 @@
     <Compile Include="UI\Components\LogicComponent.cs" />
     <Compile Include="UI\Components\RaceProviderAPI.cs" />
     <Compile Include="UI\Components\SeparatorFactory.cs" />
+    <Compile Include="UI\Components\SplitAt\SplitAtComponentFactory.cs" />
+    <Compile Include="UI\Components\SplitAt\SplitAtComponent.cs" />
+    <Compile Include="UI\Components\SplitAt\SplitAtSettings.cs">
+      <SubType>UserControl</SubType>
+    </Compile>
+    <Compile Include="UI\Components\SplitAt\SplitAtSettings.Designer.cs">
+      <DependentUpon>SplitAtSettings.cs</DependentUpon>
+    </Compile>
     <Compile Include="UI\GradientType.cs" />
     <Compile Include="UI\GraphicsCache.cs" />
     <Compile Include="UI\ImageBlur.cs" />
@@ -336,6 +345,9 @@
       <Generator>ResXFileCodeGenerator</Generator>
       <LastGenOutput>Resources.Designer.cs</LastGenOutput>
       <CustomToolNamespace>LiveSplit</CustomToolNamespace>
+    </EmbeddedResource>
+    <EmbeddedResource Include="UI\Components\SplitAt\SplitAtSettings.resx">
+      <DependentUpon>SplitAtSettings.cs</DependentUpon>
     </EmbeddedResource>
     <EmbeddedResource Include="Updates\GitInfo.resx">
       <Generator>PublicResXFileCodeGenerator</Generator>
--- a/LiveSplit/LiveSplit.Core/UI/Components/ComponentRenderer.cs
+++ b/LiveSplit/LiveSplit.Core/UI/Components/ComponentRenderer.cs
@@ -1,15 +1,29 @@
 ﻿using LiveSplit.Model;
 using LiveSplit.Options;
+using LiveSplit.UI.Components.SplitAt;
 using System;
 using System.Collections.Generic;
 using System.Drawing;
+using System.Drawing.Drawing2D;
 using System.Linq;
 
 namespace LiveSplit.UI.Components
 {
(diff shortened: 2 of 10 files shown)


DerKO9:
I haven't tried this out but I really like this feature. I love using the [MemoryGraph](https://github.com/kugelrund/LiveSplit.MemoryGraph) component and its always a little awkward trying to get both the graph and the splits into a layout that doesn't make the graph too small or the splits too big. With this, I could probably have my splits on one row and the graph on the other.

DarkRTA:
This is genuinely a good feature and something I feel like we should add to livesplit-core.

Joshimuz:
I've been using this for a few weeks now and it works perfectly, it's great for when you run out of screen space and need a lot of splits for your stream layout.

iTNTPiston:
This is a great feature. Why haven't it been added yet?

DarkRTA:
> This is a great feature. Why haven't it been added yet?

LiveSplit is at a state where the only PRs that have any chance of being accepted are bug fixes.

marcusziade:
For non-devs reading this: The reason PR's aren't being added is because all new features, no matter small or correctly implemented they seem, add complexity, which adds maintenance overhead. this is software 101. the project needs funding or someone to take over, who is also capable of maintaining the code.

Joshimuz:
Shame, seeing how this one piece of software is basically associated with all speedruns and runners.
I know it's not as simple as "just make a Patreon" but I'd hope that the community would recognise the importance of LiveSplit if such a thing did happen, simple donate button on their website aside.
I've *still* been using this since @tduva made it, with the now old version of LiveSplit it was forked from and everything still works absolutely fine lol

DarkRTA:
The reason LiveSplit has become essentially unmaintained is because we have been working on Livesplit One which has a much cleaner codebase and is practically completed in terms of core functionality.

A lot of the current work has been rewriting the entire system used for auto splitters and finding a suitable GUI framework for a native desktop port.

CryZe:
Yeah so we are actually making lots of progress with LiveSplit, just not in the original LiveSplit. The idea was to have a clean break similar to how OBS Studio and Minecraft Bedrock did it, to have a long term maintainable, testable, flexible, cross platform codebase. This has been going great, but it definitely meant that the original LiveSplit mostly only receives bug fixes. While we aren't opposed to any big new features in the original LiveSplit, they have to at least be compatible with what LiveSplit One is doing.

Personally, I'm not sure I ever even saw this issue, but this is definitely something we've been working towards with LiveSplit One as well, so we absolutely want to support this. However, as I said, it needs to be compatible with what we design there, so in this particular case, I actually had a different design in mind where you have a group of child components that get laid out in the opposite direction. So for example you could have a vertical layout, but the graph is to the left of the timer, while the rest of the components are still laid out vertically.

LiveSplit One actually recently unified its rendering across all the different frontends, so it's actually very feasible to experiment with this now and see if the "start new column / row" or "child group in a different direction" works out better. Once we finalized a design there, I'm willing to merge a compatible design over here.

marcusziade:
Good to know, sry for misguiding people, didn't know. Also, great to hear. I'll be looking into contributing if possible :)

CryZe:
Alright, so yes, I have to agree that so far the desktop versions for LiveSplit One have been mostly fruitless. We settled on Druid a while ago... and then the maintainers dropped it. So right now we don't really have anything great other than Tauri as a browser wrapper, the Druid version and an OBS plugin. So I definitely can agree that no one ever really heard of LiveSplit One if there's no real usable desktop version that we actively promote.

However, my points above about the original LiveSplit still mostly stand. We can never really make it available on Linux, macOS, the web or anything, because it's stuck on .NET Framework with Windows Forms and Microsoft kinda gave up on all of that. While we could maybe at least upgrade to the latest .NET version, that would certainly break all components and probably most ASL scripts too. And doing massive changes to any internals without even changing any .NET version would also break all components and a lot of ASLs too, because they use all sorts of internals. So essentially mostly freezing the original LiveSplit to make sure it stays working, is probably the best thing we can do.

And also neither woofer nor me have any real motivation and / or time left to really implement much of anything new for the original LiveSplit. That is unfortunately the reality.

However, there's two positive things:

1. I recently have started letting Copilot explore lots of unimplemented features, and that has really allowed me to progress much more quickly with LiveSplit One. I also have given it the task of exploring the various remaining UI frameworks that I haven't yet checked out (so far in February: WinUI 3, GTK, gpui). And that has been very fruitful as well. The GTK version is almost ready for example. So I think within the next few months we are definitely going to see actual usable Desktop versions cropping up.

This also means I can probably within the next few days also let Copilot explore the various design possibilities for the rows / columns for LiveSplit layouts. So that should at least unblock this Pull Request.

2. I also don't necessarily agree anymore that it needs to work identically in LiveSplit One. As long as it can at least somewhat migrate towards LSO, then it should be no problem to just merge this as is. And I think even if LiveSplit One ends up with a hierarchy / groups of components rather than a "switch to the next column" component, can it be migrated as just a big group at the top level.

PR #2038, opened 2020-10-19
Unittest cleanup

rpgrca:
Refactored DigitsFormatTests.cs existing test (better name, object initializer, type inference, no logic in test).
Added a few tests more to cover other cases not covered by previous one (especially handling the null case).


rpgrca:
Thanks. Any objection about switching from MSTest to Xunit? I can write tests but I prefer using Xunit instead.

wooferzfg:
@rpgrca It doesn't really matter to me which of those we use, as long as it works in the CI pipeline that we have set up with GitHub Actions.

PR #2070, opened 2021-01-10
Xunit migration

rpgrca:
Unit testing overhaul

- migrated to Xunit framework (parallel test execution, NET Core compatibility, streamlined syntax, yaml still works as tests can be run with vstest)
- removed logic from tests by splitting null cases which forced branching to new tests (rationale: tests should have cyclomatic complexity of 1: no conditionals, no loops, every line of a test must be executed every time it runs)
- split multiple asserts in different tests (rationale: unit tests should test a single thing, once something is asserted no further actions other than assertions should be executed)
- moved parametrized expected value to the end of parameter list for consistence
- used DASH constant when required
- made test names more descriptive
- used object initialization and type inference when possible
- made subject under test clear by using SUT nomenclature (which will be more useful as tests become longer)

Also used class/method pair for naming tests in a couple of cases to see how people feel about it (ie: _SegmentTimesFormatterShould.FormatTimeAsDash_WhenTimeIsNullRegardlessOfAccuracy_). If people don't object I'll be renaming *Tests to *Should for humanized test names.

Unfortunately Live Unit Testing isn't working as _Codaxy.Xlio_ is a NET 3.5 project which calls _Sgen.exe_ (temporarily adjusting Framework version to 4.6.1 reveals the error), this happened even before the Xunit migration (I'm pretty much sure this could be compiled in 4.7.2 or even Core 3.1 if not for that assembly, ever tried that?).

Didn't modify _AutoSplitterXML.cs_ as it's not a unit test, it's an integration test which should be in a different test assembly (however since Live Unit Testing is not working it's not that much of a problem right now). I'll eventually refactor the test. I can also further refactor existing tests to minimize duplication of code, and start building more tests.

Any question / suggestion welcomed.

### DIFF
--- a/LiveSplit/LiveSplit.Tests/AutoSplitterTests/AutoSplitterXML.cs
+++ b/LiveSplit/LiveSplit.Tests/AutoSplitterTests/AutoSplitterXML.cs
@@ -1,35 +1,34 @@
 ﻿using System;
-using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System.IO;
 using System.Xml;
 using System.Linq;
 using System.Collections.Generic;
-using LiveSplit.Model;
 using System.Text.RegularExpressions;
+using Xunit;
+using LiveSplit.Model;
 
 namespace LiveSplit.Tests.AutoSplitterTests
 {
-    [TestClass]
     public class AutoSplitterXML
     {
-        [TestMethod]
+        [Fact]
         public void TestAutoSplittersXML()
         {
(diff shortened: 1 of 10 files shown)


rpgrca:
Apparently deleting the whole ToolPath attribute/value from Codaxy.Xlio's SGen tag (as long as sgen.exe is in the path) fixes Live Unit Testing problem, won't commit that fix there though but encourages me to refactor AutoSplitterXML test.

PR #2071, opened 2021-01-11
Autosplitter for Super Meat Boy Forever

ldmnt:


wooferzfg:
Oops, this shouldn't have been merged.

ldmnt:
Got it, sorry for the oversight, i mechanically followed the old procedure. I’ll open a PR in the right place.

PR #2095, opened 2021-02-23
Add Refresh Rate option to settings (ROMaster2 Merge)

### LINKED ISSUE #578: Option to change LiveSplit's refresh rate
The current setting of 40Hz makes things sort of odd for streams that display 1/100 seconds. If a run is manually retimed, it has a bit of drift when compared against the recording. An option to change the refresh rate would be helpful for competitive runs or just making sure the timing is right on something. I suggest having a hard limit of 300Hz (similar to various games).

CPU usage shouldn't be a problem for most people, and if it is, they can leave it at the default value. A tooltip could appear when hovering over the input box warning against setting it too high.

### LINKED ISSUE #2058: [Feature Request] Add a way to turn down refresh rate of LiveSplit
I know wsplit has this and i think it would be a good idea to add it on livesplit I don't know how hard it would be but here is a screenshot of the feature in wsplit.
image
Its mostly so if i want to have 2 decimals on my timer it doesn't use most of the low end CPU. :)

Avasam:
This is simply an updated version of #1195 with the updated settings and merge conflicts resolved.

Original description from @ROMaster2 :
The refresh rate of the timer can now be manually adjusted between 20 and 300 hertz. 20 was arbitrarily picked, but 300 was picked as it's the maximum frame rate supported by High Efficiency Video Coding. Anything higher than that will have unnoticeable real time differences.

I did not include an option for an unlocked framerate. When testing that LiveSplit locked up anytime my mouse moved over it.

Closes #578 and Closes #2058

### DIFF
--- a/LiveSplit/LiveSplit.View/View/SettingsDialog.cs
+++ b/LiveSplit/LiveSplit.View/View/SettingsDialog.cs
@@ -53,6 +53,12 @@ namespace LiveSplit.View
             set { Settings.HotkeyProfiles[SelectedHotkeyProfile].AllowGamepadsAsHotkeys = value; }
         }
 
+        public int RefreshRate
+        {
+            get { return Settings.RefreshRate; }
+            set { Settings.RefreshRate = Math.Min(Math.Max(value, 20), 300); }
+        }
+
         public event EventHandler SumOfBestModeChanged;
 
         public string RaceViewer { get { return Settings.RaceViewer.Name; } set { Settings.RaceViewer = Web.SRL.RaceViewer.FromName(value); } }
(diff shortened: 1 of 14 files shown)


Avasam:
@wooferzfg Is there anything holding back this change?

LiveSplit/LiveSplit.View/View/SettingsDialog.cs
```diff
@@ -53,6 +53,12 @@ public bool AllowGamepadsAsHotkeys
             set { Settings.HotkeyProfiles[SelectedHotkeyProfile].AllowGamepadsAsHotkeys = value; }
         }
 
+        public int RefreshRate
+        {
+            get { return Settings.RefreshRate; }
+            set { Settings.RefreshRate = Math.Min(Math.Max(value, 20), 300); }
```
NoTeefy, line 59: Is there any reason you've used Math.Min and Math.Mix instead of Math.Clamp(value, 20, 300)?

ROMaster2: Because I didn't know about it at the time :p

Avasam: Same here. I'll update this.

wooferzfg: Looks like `Math.Clamp` isn't available on this version of .NET?

just-ero:
Has this been tested with ASL scripts?

ASL scripts use a variable called `refreshRate`, which is meant to change exactly that. Are these values connected?

PR #2116, opened 2021-04-20
Add game "Trackmania Club Campaigns"

KableKiB:
Adding Trackmania Club Campaigns for the new Trackmania leaderboards at: https://www.speedrun.com/tmclubs


KableKiB:
I found the new repo for autosplitters, will submit there instead.

PR #2137, opened 2021-06-07
Added Max Payne 3 to LiveSplit.AutoSPlitter.xml

gunlinux:
Added Max Payne 3 autosplitter to LiveSplit.AutoSPlitter.xml

Got permission from original author


wooferzfg:

PR #2138, opened 2021-06-09

CarcinogenSDA:
Trying this again, this time I fixed my stupid error.

Submitting autosplitter for Resident Evil 2 (1998, PC)


wooferzfg:

PR #2145, opened 2021-06-29
Add feature: Option to automatically open a layout when opening a split.

Relacibo:
Everytime you want to switch to another game or another category, you currently have to switch the split AND the layout. With this feature i just wanted to make this one click. 

You now have the option in the edit splits menu to select a layout to automatically open. Therefore a new Property in the Run-Class is needed: layout path, so that it will be saved in the splits XML (rust code change was also needed, i think). If the checkbox is unselected, the layoutPath-Property is null, otherwise its the file path to the layout file. With the combobox you can choose from the recently opened layouts. You can also use the browse button, to open an open file dialog. 

livesplit


CryZe:
This seems to adjust livesplit-core, can you open a Pull Request there too?

LiveSplit/LiveSplit.Core/Options/SettingsSavers/XMLSettingsSaver.cs
```diff
@@ -43,6 +43,7 @@ public void Save(ISettings settings, System.IO.Stream stream)
                 splitsFileElement.SetAttribute("categoryName", splitsFile.CategoryName);
                 splitsFileElement.SetAttribute("lastTimingMethod", splitsFile.LastTimingMethod.ToString());
                 splitsFileElement.SetAttribute("lastHotkeyProfile", splitsFile.LastHotkeyProfile.ToString());
+                splitsFileElement.SetAttribute("LayoutPath", splitsFile.LayoutPath);
```
wooferzfg, line 46: I don't think we need this change.

LiveSplit/LiveSplit.Core/Options/RecentSplitsFile.cs
```diff
@@ -20,13 +21,14 @@ public RecentSplitsFile(string path, IRun run, TimingMethod method, string hotke
...
 
-        public RecentSplitsFile(string path, TimingMethod method, string hotkeyProfile, string gameName = null, string categoryName = null)
+        public RecentSplitsFile(string path, TimingMethod method, string hotkeyProfile, string gameName = null, string categoryName = null, string layoutPath = null)
         {
             GameName = gameName;
             CategoryName = categoryName;
             Path = path;
             LastTimingMethod = method;
             LastHotkeyProfile = hotkeyProfile;
+            LayoutPath = layoutPath;
```
wooferzfg, line 31: Same thing here - I don't think we need to change this file.

LiveSplit/LiveSplit.Core/Model/IRun.cs
```diff
@@ -29,6 +29,9 @@ public interface IRun : IList<ISegment>, ICloneable, INotifyPropertyChanged
 
         bool HasChanged { get; set; }
         string FilePath { get; set; }
+        string LayoutPath { get; set; }
+
+
```
wooferzfg, line 34: We don't need these extra newlines.

LiveSplit/LiveSplit.View/View/RunEditorDialog.cs
```diff
@@ -708,6 +709,113 @@ private void picGameIcon_DoubleClick(object sender, EventArgs e)
                 }
             }
         }
+        private void btnBrowseLayout_Click(object sender, EventArgs e)
```
wooferzfg, line 712:
```suggestion

        private void btnBrowseLayout_Click(object sender, EventArgs e)
```

LiveSplit/LiveSplit.View/View/RunEditorDialog.cs
```diff
@@ -708,6 +709,113 @@ private void picGameIcon_DoubleClick(object sender, EventArgs e)
...
+        private void SaveAndApplyLayoutPath(string layoutPath)
+        {
+            Run.LayoutPath = layoutPath;
+            try
+            {
+                Application.OpenForms.OfType<TimerForm>().First().OpenLayoutFromFile(layoutPath);
+            }
+            catch (Exception exc)
+            {
+                Console.WriteLine(exc.StackTrace);
```
wooferzfg, line 794:
```suggestion
                Log.Error(exc);
```

wooferzfg (commented):
I think the layout dropdown in the Splits Editor should have a "Default" option that corresponds to the default layout.

PR #2147, opened 2021-07-06
Faster loading of edit splits menu.

Relacibo:
When I want to open the edit splits menu a second time during execution of the program, it takes nearly 4 seconds until the window is showing. With this fix, thats no longer the case. I moved the fillCbxGameName-Logic to until after the form is initialized and added a call to Application.DoEvents() to render the form before filling the combobox.


wooferzfg:
Thanks!

PR #2161, opened 2021-08-25
Compatibility w/ MacOS & Linux Through Bash

ghost:
### I found that the issue likely comes down to the process of compilation, rather than what the program uses itself.
It looks like the `.github/workflows/ci.yml` file runs some commands strictly used by Windows 10.
In addition, there are `Form` modules that seem to only run with Windows as well.
So, to work with this, I decided to give it a try to make a Bash script that follows the process.
There are already a few substitutions that will be needed to compete for compatibility, but I am sure things will eventually come together.

_I will say, though, that I will need help on working with the GitHub actions for this repository, as I am unfamiliar with how they work. Also, please let me know where the script should be moved to. Thanks._

### DIFF
--- a/LiveSplit.AutoSplitters.xml
+++ b/LiveSplit.AutoSplitters.xml
@@ -2,6 +2,28 @@
 <AutoSplitters>
     <AutoSplitter>
         <Games>
+            <Game>Blue Fire</Game>
+        </Games>
+        <URLs>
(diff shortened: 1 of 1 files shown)


wooferzfg:
GitHub Actions supports MacOS and Linux. Here's an example: https://github.com/LagoLunatic/wwrando/blob/master/.github/workflows/python-app.yml

zach83:
That is definitely good to know, @wooferzfg.
I suppose once I get a test going for compiling the project on the other platforms, I can start moving elements from the shell script to the build action file.

PR #2166, opened 2021-09-07

OCircles:
Had each of my splitters on their own repo so I'm fixing that plus adding a couple new ones


wooferzfg:
Please use the new repo: https://github.com/LiveSplit/LiveSplit.AutoSplitters/pulls

PR #2171, opened 2021-09-25
Board was split off from catext

jbzdarkid:
adding new name so the splitter works


wooferzfg:
Please use the new repository: https://github.com/LiveSplit/LiveSplit.AutoSplitters/pulls

PR #2196, opened 2021-12-13
Add "Check For Updates" option to settings

IdanZel:
This personally bugged me so I implemented it.

LiveSplit automatically checks for updates after launch. Currently, this cannot be disabled. This PR adds an option to disable this. 
See checkbox at the bottom of this window:
image

By default, this option is set to `true` (i.e. check for updates is enabled).

One thing I wasn't sure about regarding this code is where exactly should this option be checked - before adding the `Hook.GamepadHookInitialized` event handler, or inside `Hook_GamepadHookInitialized()` / `CheckForUpdates()`. The current code does the former.


ZanasoBayncuh:
Okay, I have no idea how to do this but I DESPERATELY want to. What do I need to do?

PR #2209, opened 2022-01-09
Fixed Cancel button not cancelling.

MidgeOnGithub:
If the user selects "Cancel", `SetSettings` should not be called, as it is expressly against the user's (and component developers') wishes.

As-is, if a user accesses the Component Settings dialog mid-run and only clicks Cancel, it causes buggy behavior when `SetSettings` is called.

This is counter-intuitive for the user: "Nothing should happen if I hit Cancel." And it can be tricky for component developers to debug issues caused by this, especially since the issues are not persistent. For instance, when LiveSplit is restarted, the buggy behaviors due to the erroneous `SetSettings` will not happen.

### DIFF
--- a/LiveSplit/LiveSplit.View/View/ComponentSettingsDialog.Designer.cs
+++ b/LiveSplit/LiveSplit.View/View/ComponentSettingsDialog.Designer.cs
@@ -52,7 +52,7 @@
             // btnOK
             // 
             this.btnOK.Anchor = ((System.Windows.Forms.AnchorStyles)((System.Windows.Forms.AnchorStyles.Bottom | System.Windows.Forms.AnchorStyles.Right)));
-            this.btnOK.DialogResult = System.Windows.Forms.DialogResult.Cancel;
+            this.btnOK.DialogResult = System.Windows.Forms.DialogResult.OK;
             this.btnOK.Location = new System.Drawing.Point(331, 522);
             this.btnOK.Name = "btnOK";
             this.btnOK.Size = new System.Drawing.Size(75, 23);
--- a/LiveSplit/LiveSplit.View/View/ComponentSettingsDialog.cs
+++ b/LiveSplit/LiveSplit.View/View/ComponentSettingsDialog.cs
@@ -27,7 +27,6 @@ namespace LiveSplit.View
 
         private void btnCancel_Click(object sender, EventArgs e)
         {
-            Component.SetSettings(ComponentSettings);
             DialogResult = DialogResult.Cancel;
             Close();
         }


wooferzfg:
`ComponentSettings` is the settings of the component before the dialog was opened. The component is updated live whenever anything in the settings dialog changes.
- If you hit OK, we don't need to do anything, because the component already matches the settings in the settings dialog.
- If you hit cancel, we need to revert to the previous settings from before the dialog was opened. That's what the `SetSettings` call does.

MidgeOnGithub:
>     * If you hit OK, we don't need to do anything, because the component already matches the settings in the settings dialog.
> 
>     * If you hit cancel, we need to revert to the previous settings from before the dialog was opened. That's what the `SetSettings` call does.

Ah, I believe I understand. So with this "user story":
* User starts a run.
* User decides (say, during an unskippable opening cutscene) to check their autosplitter (component) settings.
* User clicks "Edit Splits..." -> "Settings" to open the `ComponentSettingsDialog`.
* User verifies that all settings look correct; they are good to go!
* User wants to close the dialog without anything happening...
  * If they click "OK": the rest of `RunEditorDialog`'s `btnSettings_Click` runs due to the `DialogResult.OK` return.
  * If they click "Cancel": LiveSplit calls `SetSettings` because the "Cancel" logic is primarily for `ComponentSettings` reversion.

I propose that if the user exits the dialog and the "previous" and "final" `ComponentSettings` objects are the same, no work should be done.

I'm unsure of the best way to implement this. Perhaps a direct object comparison could be made, or perhaps a `settingsChanged` flag could be set somewhere. I will look into it, but I am open to suggestions. I believe this is important for the sanity of users and component developers alike, even if it seems like a weird edge case. (:

MidgeOnGithub:
Follow-up to my previously edited comment. After more research, I discovered that I had misidentified which Dialog's "Cancel" was the problem, and thus I had misplaced my focus for this branch's work. I investigated deeper and found an actual answer to the buggy behavior.

It stems from in `TimerForm`'s, `EditSplits()`. Here, the `RunEditorDialog` is created. And `if (result == DialogResult.Cancel)` eventually `SetRun(runCopy)` is called, _regardless of if `Run` differs from `runCopy`_. `SetRun` deactivates and recreates the autosplitting component.

If this happens mid-run, it causes the weird behavior I originally described because everything within the developed Component/Autosplitter has been re-initialized, so all sense of "state" has been erased.

MidgeOnGithub:
Given the new information provided by my more recent comments and edits, it is clear to me that the original focus of this branch was inaccurate. My apologies.

At some point in the future, I might try resolving the `RunEditorDialog` "Cancel" issue I described above. But for the time being, I don't believe this PR is applicable or active enough to remain open.

I created #2214 to document the issue.

PR #2211, opened 2022-01-12
"Submit run to speedrun.com" Button in context menu

eltharynd:
So... the share to speedrun.com feature has never worked for me.. 

So i implemented my own.. 
You have to provide your API key in the Additional Info tab, then on the right click context menu you can chose submit to sr.com to automatically post the run with a placeholder video.

c# and .net are not exactly my forte so if you guys could have a look at it that would be great.. 
tyvm


wooferzfg:
Would it be possible to fix the issues with the existing functionality instead?

eltharynd:
I guess it would but i'm not even sure what they were actually intended to do.. 
Cause the name of the buttons did not represent what was in the code... 

I actually wanted to remove those since (according to google) nobody has managed to make those work for years...
But i wanted to leave this decision to you guys..

I might look into those tomorrow and see if i get it

eltharynd:
In all honesty.. 

- The "associate with speedrun.com" button doesn't make sense at all... Especially since it's in the splits options and not on a run per run basis... I would remove that
- The "submit run" can't work at all cause you need to provide an API key for that... Besides it doesn't make any sense that that would be in the splits options...

I might spend some more time on it and delete those two buttons and related functionality and polish my solution a bit better... (maybe by moving the API key in global settings for example) 
What do you folks think?

wooferzfg:
- The "associate with speedrun.com" button only appears if you have a new PB. Once the run is associated, it'll change to "view on speedrun.com" until you get a new PB. Because of that, it is per-run.
- When you click on the button for the first time, it should open a window that tries to retrieve an API key. (Not sure if that's still working?)

eltharynd:
Currently submit run opens a window that tells you you need to add an API key (but there's no way to add an API key anywhere...

Meanwhile associate run opens a popup that asks for the run url, but it accepts no  url whatsoever.. i tried with many combinations... Googling around it appears it actually is supposed to work with splits.io or something.. but that's now a paid service and, frankly i never used it and o don't know many people who use it in my community...

But now that that i understand better what the original thought behind those was i will look into it more and fix it..

Imo tho the "submit" button should not be in the edit splits/additional info.. that's hidden away behind too many clicks... I wanna sett hat stuff up once when i set my splits up then just right click and submit from the context menu...

I will remove the old buttons and their (not working) functionality and make the submit function a bit better.
This button already checks if the run is ready to be uploaded, i will also grey the option out, but i don't see reason to disable it if it's not a PB... One might still want to upload their run for statistics, or maybe they didn't get a PB on their category but it's a PB for another (games with deathless categories for example), I'd still leave that decision to the user personally...

Might i ask, while I'm at it, some clarification regards the difference between game time and time with no loads? Is it me or they're one and the same for livesplit? All i could find was the real time and another variable that stores the load time, so one minus the other is time w/o loads, but isn't that the same as in-game time too? Is there some documentation about that somewhere?

wooferzfg:
Sharing a run to speedrun.com previously opened a web page on speedrun.com that showed the API key, which LiveSplit then retrieved. If this isn't working anymore, we should update it to make it work.

"Associate run" should accept a speedrun.com run URL.

You should be able to submit a run from either "Edit Splits" or the "Share" dialog. In order to get a PR merged, I'd suggest not trying to change this behavior.

IGT and Time Without Loads are the same in LiveSplit, but distinct on Speedrun.com. It depends on which timing methods the game accepts in its rules.

eltharynd:
I don't see how you can make an automatic API key retrieval tho...
It's obviously hidden behind authentication.

wooferzfg:
Rather than a new field in the settings, can we just replace the API key retrieval that we currently have with a dialog window that has an input field for the API key? (Should be similar to the window that opens when you click "Open Splits from URL".)

wooferzfg:
The Speedrun.com API key should be in the [WebCredentials class](https://github.com/LiveSplit/LiveSplit/blob/b893917f51245bfc5dfbc890114a32de76cfa0c8/LiveSplit/LiveSplit.Core/Web/WebCredentials.cs), not in the Settings.

LiveSplit/LiveSplit.View/View/TimerForm.cs
```diff
@@ -2622,6 +2623,170 @@ private void shareMenuItem_Click(object sender, EventArgs e)
...
+                        ""verified"": false,
+                        ""times"": {{
+                                        ""realtime"": {realtime},
+                                        ""realtime_noloads"": {realtime_noloads},
+                                        ""ingame"": {ingame}
+                        }},
+                        ""comment"": ""Automatically submitted by LiveSplit, video might be temporary."",
+                        ""video"": ""https://youtube.com/watch?v=mumblefoo""
+                        VARIABLES_PLACEHOLDER
+                    }}
```
wooferzfg, line 2670: We already have SpeedrunComSharp for all of this. Please limit the scope of this PR to just replacing the current way ofbretrieving the API key (opening a webpage and grabbing it automatically) with a new way (a dialog box with a link to the settings page and an input field for the API key) - otherwise, this won't get merged.

eltharynd:
There is no "straight to API key" link until you've already logged in or inserted the username... besides... 
- I strongly disagree with the idea of having the user log in inside the app
- Especially if it's ultimately just to still copy and paste the API key

It would not only be a security risk... I will no develop something that lets user insert their own password for someone else's access not through their own API... that is just bad practice... and a safety risk... 
But even if we went beyond that: it's ultimately an annoyance... Everybody is already logged in to speedrun.com in their browser, in which they have access to many a way to make inserting the password even more secure...

I can fix the rest of the code if you want... 
But i will replace the login part with just pasting the API key inside the settings.. that way they will be available for all splits/games/cats they have... As I had already changed, btw...

eltharynd:
moved

PR #2212, opened 2022-01-14
Add setting for drop decimal poss. time save in prev. seg component

submerged-castle:
Adds support to save the drop decimal setting for possible time save in previous segment component in https://github.com/LiveSplit/LiveSplit.PreviousSegment/pull/5


wooferzfg:
Why can't you just set `DropDecimals` on the formatter instead of this new property?

PR #2213, opened 2022-01-15
Fixed Speedrun.com Authentication

eltharynd:
As discussed here 
#2211 

I added the API key field to settings
Settings

And these to use that instead
MetaData
Share


wooferzfg:
As I mentioned in [this comment](https://github.com/LiveSplit/LiveSplit/pull/2211#issuecomment-1012277772), I think it would be better and more consistent if this was a dialog box that asked you for your Speedrun.com API key rather than a field in the Settings.

Also, as I mentioned in [this comment](https://github.com/LiveSplit/LiveSplit/pull/2211#issuecomment-1012309081), this should be in the WebCredentials instead of the Settings.

I did not suggest having the user log in inside the app - I suggested having a dialog window that pops up that **links to your settings page** and **has an input field for your API key**. You can think of this as taking your current implementation and moving it from the Settings to a dialog window that pops up. You should be replacing the code that is in [SpeedrunComOAuthForm](https://github.com/LiveSplit/LiveSplit/blob/master/LiveSplit/LiveSplit.View/View/SpeedrunComOAuthForm.cs).

wooferzfg:
I updated the Speedrun.com authentication process here: https://github.com/LiveSplit/LiveSplit/commit/4235dda32e752b9b5a833572be0c5c282a16054b

PR #2226, opened 2022-02-15
Add Eye of the Temple to AutoSplitters

runevision:
I'm the developer of Eye of the Temple and I made this simple autosplitter based on the in-game timer.

I don't know if or how speedrunners might want to split their runs so it only splits when the timer stops for now. However, there's info in the ASL file that others could use to extend its functionality if they want.


wooferzfg:
Please update this repo instead: https://github.com/LiveSplit/LiveSplit.AutoSplitters

PR #2236, opened 2022-03-19
Adjust length of sleep instruction on TimerWorker for time spent refreshing

SHTDJ:
This will increase the consistency of the timer refresh rate when more resource intensive components are used (ie load removers using optical character recognition)

On a side note: Due to delay between a thread's sleep duration ending and the thread actually resuming execution the refresh rate entered in the settings will differ vastly from the refresh rate actually achieved (ie entering a refresh rate of 30 will result in an actual refresh rate of 23)

### DIFF
--- a/LiveSplit/LiveSplit.View/View/TimerForm.cs
+++ b/LiveSplit/LiveSplit.View/View/TimerForm.cs
@@ -33,7 +33,6 @@ using System.Net;
 using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading;
-using System.Timers;
 using System.Threading.Tasks;
 using System.Windows.Forms;
 using System.Xml;
@@ -79,8 +78,6 @@ namespace LiveSplit.View
         protected float TotalPosition { get; set; }
 
         private bool DontRedraw = false;
-
-        private StringBuilder sb = new StringBuilder();
 
         protected Region UpdateRegion { get; set; }
 
@@ -1179,9 +1176,6 @@ namespace LiveSplit.View
 
         void TimerElapsed()
         {
-            //((System.Timers.Timer)source).Interval = 10;
-            //int logCount = 0;
-            decimal updateStartTime = DateTime.Now.Ticks;
             try
             {
                 this.InvokeIfRequired(() =>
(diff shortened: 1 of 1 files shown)


wooferzfg:
> entering a refresh rate of 30 will result in an actual refresh rate of 23

Is there a way to resolve this?

goofball564:
Sorry for the minor wall of text.

#### Re: Refresh Rate Inaccuracy
One solution to inaccurate refresh rates would be calls to [`timeBeginPeriod`](https://docs.microsoft.com/en-us/windows/win32/api/timeapi/nf-timeapi-timebeginperiod) and [`timeEndPeriod`](https://docs.microsoft.com/en-us/windows/win32/api/timeapi/nf-timeapi-timeendperiod). This affects a global Windows setting and comes with significantly increased power consumption, reduced battery life, and reduced system performance, since it means Windows's thread scheduler is awoken up to 15.625x more often than normal. Otherwise Windows's default system timer resolution of 64 ticks/second, which limits the precision of Thread.Sleep to 15.625 milliseconds.

#### Re: `DateTime.Now.Ticks`
Wouldn't it be better to use a [`Stopwatch`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch) than `DateTime.Now.Ticks`? It's (probably) much more precise, and measuring elapsed time is what `Stopwatch` is meant for. [`DateTime.UtcNow`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch)`.Ticks` would at least be less expensive since it doesn't have to do time zone calculations. It's also `Stopwatch`'s backup if `QueryPerformanceCounter` isn't available.

`DateTime.UtcNow.Ticks` and `DateTime.Now.Ticks` are affected by the system clock being changed and are limited in accuracy by Windows's system timer resolution. `DateTime.Now` could also maybe potentially risk time zone related bugs.

thekovic:
> This affects a global Windows setting and comes with significantly increased power consumption, reduced battery life, and reduced system performance, since it means Windows's thread scheduler is awoken up to 15.625x more often than normal.

That sounds awful. I'd never want LiveSplit to significantly affect system performance.

SHTDJ:
Stopwatch has the exact same issue

PR #2241, opened 2022-04-12
Make UpdateManager set download destination.

### LINKED ISSUE #2240: UpdateManager's download behavior is unfriendly to component DLL authors.
I believe all component DLLs must be downloaded into the Components folder. Assuming that is correct, why does `UpdateManager` "trickle down" a component's download location based on `UpdateURL`, implicitly enforcing a rigid repo/URL structure for all component authors? Why doesn't `UpdateManager` set the destination according to LiveSplit's requirements? Component authors should not have their updates fail after they successfully fulfill their apparent end of the contract by setting `UpdateURL` + `FileChange.Path`s as well as `XMLURL` accurately. In other words, an author's repo/URL structure should not be dependent on LiveSplit's choice of the Components folder nor `UpdateManager`'s lack of actual work in setting the destination file name.

Currently, `UpdateManager` uses (`IUpdateable.UpdateURL` + `FileChange.Path`) as the download URL, which is what an author could intuitively expect. But then it assumes that a basic `string.replace()` on `FileChange.Path` yields the correct destination, despite the fact that LiveSplit seems to absolutely require that component DLLs be in the Components folder. This "minimal" implementation implicitly dictates the directory structures of `IUpdateable`'s `UpdateURL` and `XMLURL` as well as `FileChange.Path`s. 

A basic example of when this causes an unfriendly failure is if an author sets their `IComponentFactory` (thus, `IUpdateable`) implementation's `UpdateURL` to be a directory which is not the direct parent of a Components folder/url. When the XML from `XMLURL` is parsed, the `<file>` elements will contain `path`s which do not directly point at the Components folder, which is implicitly required by `UpdateManager`'s current behavior.

Sample:
> IComponentFactory:
> 
> ```cs
> ...
> UpdateURL = "https://raw.githubusercontent.com/.../.../main/";
> XMLURL = UpdateURL + "SomeGame/Components/update.xml";
> ...
> ```
> 
> Update XML:
> ```cs
> ...
> <files>
>   <file path="SomeGame/Components/SomeGame.dll" status="changed"/>
> </files>
> ...
> ```
> 
> Required Target Destination: `<LiveSplit>/Components/SomeGame.dll`
> Actual Destination: `<LiveSplit>/SomeGame/Components/SomeGame.dll`

Here, `SomeGame.dll` would be fetched successfully but placed in the wrong directory. Then, LiveSplit would be unable to find the newly downloaded file and thus `UpdateManager` would be repeatedly called to update the component while it continually places the DLL incorrectly.

MidgeOnGithub:
Closes #2240.

### DIFF
--- a/LiveSplit/UpdateManager/Updater.cs
+++ b/LiveSplit/UpdateManager/Updater.cs
@@ -101,8 +101,14 @@ namespace UpdateManager
 
             public void PerformUpdate()
             {
-                string GetLocalComponentsPath(string xmlChangePath) =>
-                    "Components" + Path.DirectorySeparatorChar + xmlChangePath.Split('/').Last();
+                string ConvertChangeUrlPartToPath(string xmlChangePath)
+                {
+                    string path = xmlChangePath.Replace('/', Path.DirectorySeparatorChar);
+                    bool thirdPartyDll = !UpdateURL.Equals("http://livesplit.org/update/");
+                    return thirdPartyDll
+                        ? Path.Combine("Components", path.Split(Path.DirectorySeparatorChar).Last())
+                        : path;
+                }
 
                 IList<Update> updates = Updates.Where(x => x.Version > Version).ToList();
                 List<string> addedFiles = new List<string>();
@@ -143,15 +149,15 @@ namespace UpdateManager
                 int fileChangesCount = addedFiles.Concat(changedFiles).Concat(removedFiles).Count();
                 double i = 0;
 
-                foreach (string path in addedFiles.Concat(changedFiles))
+                foreach (string xmlChangePath in addedFiles.Concat(changedFiles))
                 {
-                    DownloadFile(UpdateURL + path, GetLocalComponentsPath(path));
+                    DownloadFile(UpdateURL + xmlChangePath, ConvertChangeUrlPartToPath(xmlChangePath));
                     UpdatePercentageRefreshed?.Invoke(this, new UpdatePercentageRefreshedEventArgs(++i / fileChangesCount));
                 }
 
-                foreach (string path in removedFiles)
+                foreach (string xmlChangePath in removedFiles)
                 {
-                    File.Delete(GetLocalComponentsPath(path));
+                    File.Delete(ConvertChangeUrlPartToPath(xmlChangePath));
                     UpdatePercentageRefreshed?.Invoke(this, new UpdatePercentageRefreshedEventArgs(++i / fileChangesCount));
                 }
             }
--- a/LiveSplit/UpdateManager/Updater.cs
+++ b/LiveSplit/UpdateManager/Updater.cs
@@ -101,6 +101,9 @@ namespace UpdateManager
 
             public void PerformUpdate()
             {
+                string GetLocalComponentsPath(string xmlChangePath) =>
+                    "Components" + Path.DirectorySeparatorChar + xmlChangePath.Split('/').Last();
+
                 IList<Update> updates = Updates.Where(x => x.Version > Version).ToList();
                 List<string> addedFiles = new List<string>();
                 List<string> changedFiles = new List<string>();
@@ -142,13 +145,13 @@ namespace UpdateManager
 
                 foreach (string path in addedFiles.Concat(changedFiles))
                 {
-                    DownloadFile(UpdateURL + path, path.Replace('/', '\\'));
+                    DownloadFile(UpdateURL + path, GetLocalComponentsPath(path));
                     UpdatePercentageRefreshed?.Invoke(this, new UpdatePercentageRefreshedEventArgs(++i / fileChangesCount));
                 }
 
                 foreach (string path in removedFiles)
                 {
-                    File.Delete(path);
+                    File.Delete(GetLocalComponentsPath(path));
                     UpdatePercentageRefreshed?.Invoke(this, new UpdatePercentageRefreshedEventArgs(++i / fileChangesCount));
                 }
             }
--- a/LiveSplit/UpdateManager/Updater.cs
+++ b/LiveSplit/UpdateManager/Updater.cs
@@ -101,47 +101,31 @@ namespace UpdateManager
 
             public void PerformUpdate()
             {
-                string ConvertChangeUrlPartToPath(string xmlChangePath)
-                {
-                    string path = xmlChangePath.Replace('/', Path.DirectorySeparatorChar);
-                    bool thirdPartyDll = !UpdateURL.Equals("http://livesplit.org/update/");
-                    return thirdPartyDll
-                        ? Path.Combine("Components", path.Split(Path.DirectorySeparatorChar).Last())
-                        : path;
-                }
+                string ConvertChangeUrlPartToPath(string xmlChangePath) => xmlChangePath.Replace('/', Path.DirectorySeparatorChar);
 
                 IList<Update> updates = Updates.Where(x => x.Version > Version).ToList();
-                List<string> addedFiles = new List<string>();
-                List<string> changedFiles = new List<string>();
-                List<string> removedFiles = new List<string>();
+                var addedFiles = new Dictionary<string, string>();
+                var changedFiles = new Dictionary<string, string>();
+                var removedFiles = new Dictionary<string, string>();
 
-                foreach (FileChange change in updates.SelectMany(x => x.FileChanges))
+                foreach (var change in updates.SelectMany(x => x.FileChanges))
                 {
-                    bool inAddedFiles = addedFiles.Contains(change.Path);
-                    bool inChangedFiles = changedFiles.Contains(change.Path);
-                    bool inRemovedFiles = removedFiles.Contains(change.Path);
-
                     string path = change.Path;
+                    string localPath = change.LocalPath;
 
                     switch (change.Status)
                     {
                         case ChangeStatus.Added:
-                            if (!inAddedFiles)
-                                addedFiles.Add(path);
-                            if (inRemovedFiles)
-                                removedFiles.Remove(path);
+                            removedFiles.Remove(path);
+                            addedFiles[path] = localPath;
                             break;
                         case ChangeStatus.Changed:
-                            if (!inChangedFiles)
-                                changedFiles.Add(path);
+                            changedFiles[path] = localPath;
                             break;
                         case ChangeStatus.Removed:
-                            if (inAddedFiles)
-                                addedFiles.Remove(path);
-                            if (inChangedFiles)
-                                changedFiles.Remove(path);
-                            if (!inRemovedFiles)
-                                removedFiles.Add(path);
+                            addedFiles.Remove(path);
+                            changedFiles.Remove(path);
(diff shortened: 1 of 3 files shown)


MidgeOnGithub:
Hopefully it isn't bothersome that I ask:
Is there anything I can do to improve the PR/Issue and/or help progress the review?

I understand there are other PRs/Issues, so of course I do not expect immediate response. And I am wondering if there is a way I can know other PRs which might be "ahead in queue", so that I can relay relative progress to my component consumers who are currently affected by this PR/Issue.

Also, I wonder if there are other ways I could help.

wooferzfg:
If you're able to test out your changes and describe the results of your testing here, that would give me the confidence to merge your changes.

MidgeOnGithub:
(Apologies for the force push... git repo mismanagement on my part.)

A critical point of concern that I was unsure how to test:

- I do not know if DLLs in the "root" of LiveSplit's installation folder (e.g., `LiveSplit.View.dll` & `SpeedrunComSharp.dll`) also update through UpdateManager's `UpdateAll`.
  - If so, then this PR's changes would download them to the wrong location (they'd be downloaded into _Components_).

How I tested with VS 2022 Preview:
- Placed an outdated DLL which contains the sort of `UpdateURL`, `XMLURL`, and upstream `update.xml` which would be problematic for the existing UpdateManager in the _Components_ folder.
  - My example:
    - I used version 1.0.0 (current: 2.0.0) of [TR4.dll](https://github.com/TombRunners/autosplitters/raw/b0c4688e34d036879011eef15e3e15e4689e85fb/TombRaiderIV/Components/TR4.dll), the autosplitter for Tomb Raider IV: The Last Revelation.
    - Here is the upstream [update.xml](https://github.com/TombRunners/autosplitters/blob/ba4f2441363b05ef07788a36c52baea56f827282/TombRaiderIV/Components/update.xml).
- Switched the _Output type_ of UpdateManager CS Project to _Console Application_, then used VS to build an UpdateManager.exe with this PR's changes.
- Placed a breakpoint at line 177 of `Updater.cs`. 
```cs
Process.Start("UpdateManager.exe", arguments);
```
(This is important because in line 175, the newest UpdateManager.exe is downloaded, and I need to ensure to test the newer one.)
- Debugged LiveSplit with VS;
  - Once breakpoint was reached, I overwrote the just-downloaded UpdateManager with the one I built.
  - I could not determine how to attach/debug the UpdateManager `Process` once programmatically launched, but I observed:
    - The "old" UpdateManager placed the file in `<LiveSplit>/TombRaiderIV/Components/TR4.dll`.
    - The "new" UpdateManager placed the file in `<LiveSplit>/Components/TR4.dll`.

MidgeOnGithub:
@wooferzfg Sorry to bother with a ping. I was curious to know if the concern in my last comment was perhaps the blocker.

Should I put some more time looking at code and/or figuring out a way to adjust or test for the "base" DLLs?

wooferzfg:
Sorry for the late response. Yes, those files are downloaded through `UpdateAll`. The `Updateable` instance for the main LiveSplit files is here: https://github.com/LiveSplit/LiveSplit/blob/b893917f51245bfc5dfbc890114a32de76cfa0c8/LiveSplit/LiveSplit.View/Updates/LiveSplitUpdateable.cs

MidgeOnGithub:
I added a check to see if the `UpdateUrl` being processed matches the one in `LiveSplitUpdateable`'s. 

If not (meaning the DLL being downloaded is not a "native" LiveSplit DLL; i.e., third-party):
The DLL will be routed to `<LiveSplit>/Components`.

Otherwise (the DLL is a "native" DLL):
The DLL's partial URL "path" is used as the desired download path (relative to LiveSplit's root dir), matching original behavior.

MidgeOnGithub:
In regards to testing:
- I used the methodology described previously to test that the latest changes function correctly for third-party DLLs.
- I do not know of a way to grab an old LiveSplit installation to test that "native" DLL updates succeed.

wooferzfg:
You can download old releases on GitHub: https://github.com/LiveSplit/LiveSplit/releases

You might be able to try loading an old build, but replacing its `UpdateManager.dll` with one that has your changes?

MidgeOnGithub:
I used [version 1.5](https://github.com/LiveSplit/LiveSplit/releases/download/1.5/LiveSplit_1.5.zip) for my test. Before using the described test methodology, I overwrote version 1.5's `UpdateManager.dll` with the newest from this branch, as suggested.
- In the "first wave" of updates, it ignored `TR4.dll` and listed "native" components for updates; all were updated successfully.
- UpdateManager popped up a second time with only `TR4.dll` listed for update; it was updated correctly (in _Components_).

In summary, where `<LiveSplit>` is the root folder LiveSplit is running from:
- DLLs downloaded from `http://livesplit.org/update/` are placed per the `FileChange`'s `Path`; e.g., `<LiveSplit>/<Path>`.
- DLLs downloaded from anywhere else will be placed in `<LiveSplit>/Components`.

MidgeOnGithub:
The only scenario I see where this "breaks" existing components is if some third-party component(s) were built with a dependency on using UpdateManager to download files into sub-directories inside of _Components_.

I'm unsure why a component would have a necessary dependency on using UpdateManager to download other DLLs into specific sub-directories. But it follows the consideration of this implementation: third-party DLLs will _only_  be downloaded into _Components_.

So long as that is an acceptable reality, I believe the state of this PR is satisfactory.

wooferzfg:
Would it make sense to support a separate optional `filePath` attribute in the XML instead of hardcoding the download path for third party components to the `Components` folder?

So if `filePath` is provided, you use that to determine where to download the DLL. Otherwise, you use `path`.

MidgeOnGithub:
> Would it make sense to support a separate optional `localPath` attribute in the XML instead of hardcoding the download path for third party components to the `Components` folder?
> 
> So if `localPath` is provided, you use that to determine where to download the DLL. Otherwise, you use `path`. That would also make these changes safer, since they would only affect components that provide a `localPath`.

I see your point. Optionally, if we wanted to not change the XML, we could change the behavior of this implementation:
If a third-party DLL's `Path` attribute contains the string "Components", we can instead use the `Path` starting from that string as the download path, instead of hardcoding the file into the _Components_ folder.

Example third-party `Path`s and logic:
1. `path="TombRaiderIV/Components/TR4.dll"`
- Find "Components" string
- Take `Substring` starting at "Components" -> `Components/TR4.dll`
- Final download location: `<LiveSplit>/Components/TR4.dll`.

2. `path="TombRaiderIV/TR4.dll"` 
- No "Components" string found 
- Assign download path using `Path`: `Path.Combine("Components", "TR4.dll")`
- Final download location: `<LiveSplit>/Components/TR4.dll`.

3. `path="TombRaiderIV/Components/LibraryOrSomething/SomeDLL.dll"`
- Find "Components" string
- Take `Substring` starting at "Components" -> `Components/LibraryOrSomething/SomeDLL.dll`
- Final download location: `<LiveSplit>/Components/LibraryOrSomething/SomeDLL.dll`

@wooferzfg Any preference between your suggestion and this proposal?

MidgeOnGithub:
An addendum to my last comment, the proposed change could look like this:
```cs
string ConvertChangeUrlPartToPath(string xmlChangePath)
{
    string path = xmlChangePath.Replace('/', Path.DirectorySeparatorChar);

    // LiveSplit Components
    if (UpdateURL.Equals("http://livesplit.org/update/")) 
        return path;
    
    // Third-party Components
    const string componentsString = "Components";
    int index = path.IndexOf(componentsString, StringComparison.InvariantCultureIgnoreCase);
    return index == -1
        ? Path.Combine(componentsString, path.Split(Path.DirectorySeparatorChar).Last())
        : path.Substring(index);
}
```

wooferzfg:
I would prefer just having a new optional attribute because it's guaranteed not to break any existing components, and it's a bit more explicit about its behavior.

MidgeOnGithub:
I added the XML attribute as `localPath` and the respective `FileChange` member being `LocalPath`. I made it optional by adding a conditional accessor to `node.Attributes["localPath"]`. If it is not null, it will be used to determine the download path.

MidgeOnGithub:
@wooferzfg Bump 😄

LiveSplit/UpdateManager/Updater.cs
```diff
@@ -101,54 +101,51 @@ public IEnumerable<string> GetChangeLog()
...
+                    string localPath = change.LocalPath;
 
                     switch (change.Status)
                     {
                         case ChangeStatus.Added:
-                            if (!inAddedFiles)
-                                addedFiles.Add(path);
-                            if (inRemovedFiles)
-                                removedFiles.Remove(path);
+                            removedFiles.Remove(path);
```
wooferzfg, line 119: What's the reasoning for changing these lines?

MidgeOnGithub: I read that [Dictionary.Remove](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.remove) returns `false` if `key` is not found. And since the dictionary was initialized with `new` above, a null exception should not occur. So I removed the safeguarding `if`s.
No problem if you would like these changes reverted.

LiveSplit/UpdateManager/Updater.cs
```diff
@@ -101,54 +101,51 @@ public IEnumerable<string> GetChangeLog()
...
 
-                foreach (FileChange change in updates.SelectMany(x => x.FileChanges))
+                foreach (var change in updates.SelectMany(x => x.FileChanges))
                 {
-                    bool inAddedFiles = addedFiles.Contains(change.Path);
-                    bool inChangedFiles = changedFiles.Contains(change.Path);
-                    bool inRemovedFiles = removedFiles.Contains(change.Path);
-
                     string path = change.Path;
+                    string localPath = change.LocalPath;
```
wooferzfg, line 114: I think you can do the defaulting here, rather than in two places when adding or removing files.
```suggestion
                    string localPath = change.LocalPath ?? change.Path;
```

wooferzfg:
I think this is safe to merge if you can confirm that you've tested your latest changes here with both the traditional LiveSplit updates and your own components.

MidgeOnGithub:
@wooferzfg I've had a lot of real life events preventing me from testing again. Would it be possible for you to test, perhaps using the methodology I described (or another you find appropriate)?

I think this is a great feature but I am indefinitely unavailable. I can try to make some time if you are unable, but I can't say when that would be.

MidgeOnGithub:
Commenting to say that I haven't forgotten about this, hoping it isn't considered a stale PR. I will ask and see if I know anyone who could perform the last testing (hopefully) for me.

MidgeOnGithub:
I finally got the ability to re-test with my latest changes.

I can confirm that my testing methodology succeeds with the latest changes for an older version of LiveSplit (such as 1.5) and the latest version.

@wooferzfg I believe this is safe to merge. Given there has been a good bit of time since the last time we discussed this, do you feel there is anything else that should happen that I would need to do before this is merged?

PR #2248, opened 2022-05-22
Added series title to list of tokenized strings for Game Search, this…

pugpugpugs:
… allows the matching to be more accurate.

Removed duplicate code.

### DIFF
--- a/LiveSplit/LiveSplit.Core/Model/AbbreviationExtensions.cs
+++ b/LiveSplit/LiveSplit.Core/Model/AbbreviationExtensions.cs
@@ -41,6 +41,7 @@ namespace LiveSplit.Model
                     list.AddRange(seriesTitleAbbreviations);
                 }
                 list.AddRange(subTitleAbbreviations);
+                list.Add(seriesTitleTrimmed);
 
                 var isThereOnlyOneSeriesTitleAbbreviation = seriesTitleAbbreviations.Count == 1;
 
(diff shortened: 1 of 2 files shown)


CryZe:
If I recall correctly these abbreviations are also used as the abbreviations for the title component when the game name doesn't fit. So showing just the series there is not what we want. For game search it would make sense though.

pugpugpugs:
Do you have an example of that scenario so I can test/look into it? By my testing ,this seems to work.

CryZe:
Doesn't this also add just `Alien` to the list of possible titles the title component might show?

pugpugpugs:
Maybe the issue is that component is not working as it was intended in the first place?

`                if (Settings.SingleLine && Settings.ShowGameName && Settings.ShowCategoryName)
                {
                    var text = string.Format("{0} - {1}", state.Run.GameName, extendedCategoryName);
                    var gameAbbreviations = state.Run.GameName.GetAbbreviations();
                    var shortestGameName = gameAbbreviations.Last();
                    var categoryAbbreviations = getCategoryNameAbbreviations(extendedCategoryName);
                    var combinedAbbreviations1 = gameAbbreviations.Select(x => string.Format("{0} - {1}", x, extendedCategoryName));
                    var combinedAbbreviations2 = categoryAbbreviations.Select(x => string.Format("{0} - {1}", shortestGameName, x));
                    var abbreviations = combinedAbbreviations1.Concat(combinedAbbreviations2).ToList();
                    GameNameLabel.Text = text;
                    GameNameLabel.AlternateText = mode == LayoutMode.Vertical ? abbreviations : new List<string>();
                    CategoryNameLabel.Text = "";
                }`

PR #2280, opened 2022-11-02
Add support for ScriptType element in AutoSplitters XML

wooferzfg:
This PR updates the AutoSplitters XML format so support a new `ScriptType` element for AutoSplittingRuntime scripts. To define an Auto Splitter like this, you would specify:
```
<Type>Script</Type>
<ScriptType>AutoSplittingRuntime</ScriptType>
```
For all other Auto Splitters, the format isn't changed.

I also verified that LiveSplit 1.8.21 (before we added the new AutoSplittingRuntime component) loads even with an Auto Splitter defined in this new format.

### DIFF
--- a/LiveSplit/LiveSplit.Tests/AutoSplitterTests/AutoSplitterXML.cs
+++ b/LiveSplit/LiveSplit.Tests/AutoSplitterTests/AutoSplitterXML.cs
@@ -30,15 +30,10 @@ namespace LiveSplit.Tests.AutoSplitterTests
             var urlElems = autoSplitterElems.SelectMany(x => x["URLs"].ChildNodes.OfType<XmlElement>());
             Assert.True(urlElems.All(x => x.Name == "URL"), "<URLs> must contain only <URL> elements");
 
-            IDictionary<string, AutoSplitter> autoSplitters = autoSplitterElems.Select(element =>
-                    new AutoSplitter()
-                    {
-                        Description = element["Description"].InnerText,
-                        URLs = element["URLs"].ChildNodes.OfType<XmlElement>().Select(x => x.InnerText).ToList(),
-                        Type = (AutoSplitterType)Enum.Parse(typeof(AutoSplitterType), element["Type"].InnerText),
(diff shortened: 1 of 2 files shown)


LiveSplit/LiveSplit.Tests/AutoSplitterTests/AutoSplitterXML.cs
```diff
@@ -30,24 +30,25 @@ public void TestAutoSplittersXML()
...
             Assert.True(!autoSplitters.Values.Any(x => x.URLs.Any(y => y.EndsWith(".asl")) && x.Type == AutoSplitterType.Component),
                 "ASL Script is downloaded even though Type \"Component\" is specified");
-            Assert.True(!autoSplitters.Values.Any(x => x.URLs.First().EndsWith(".dll") && x.Type == AutoSplitterType.Script),
+            Assert.True(!autoSplitters.Values.Any(x => x.URLs.Any(y => y.EndsWith(".asl")) && x.Type == AutoSplitterType.AutoSplittingRuntimeScript),
+                "ASL Script is downloaded even though ScriptType \"AutoSplittingRuntime\" is specified");
+            Assert.True(!autoSplitters.Values.Any(x => x.URLs.First().EndsWith(".dll") && (x.Type == AutoSplitterType.Script || x.Type == AutoSplitterType.AutoSplittingRuntimeScript)),
                 "Component is downloaded even though Type \"Script\" is specified");
+            Assert.True(!autoSplitters.Values.Any(x => x.URLs.Any(y => y.EndsWith(".wasm")) && x.Type == AutoSplitterType.Component),
+                "WebAssembly Script is downloaded even though Type \"Component\" is specified");
```
wooferzfg, line 49: Not sure if there's a case where a "Component" auto splitter might want a WebAssembly script? We currently don't have any auto splitters like that though.

CryZe: It's probably fine to error out there for now.

CryZe (approved):

PR #2283, opened 2022-11-04
Fix race condition in SetGameTime

wooferzfg:
Previously, the value of `GameTimePauseTime` would take into account two separate calls to `TimeStamp.Now`, which resulted in a race condition. Now, the value is always set directly to `gameTime`.

### DIFF
--- a/LiveSplit/LiveSplit.Core/Model/LiveSplitState.cs
+++ b/LiveSplit/LiveSplit.Core/Model/LiveSplitState.cs
@@ -202,9 +202,14 @@ namespace LiveSplit.Model
 
         public void SetGameTime(TimeSpan? gameTime)
         {
-            if (CurrentTime.RealTime.HasValue && gameTime.HasValue)
+            var currentTime = CurrentTime;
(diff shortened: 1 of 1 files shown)


CryZe (approved):
It's fine if you are indeed sure that setting GameTimePauseTime doesn't mess with anything.

wooferzfg:
It definitely doesn't in our codebase. But I'm not sure if there's a random auto splitter out there using it for some weird reason. We could go back to checking `if (IsGameTimePaused)` if we want to be extra safe.

wooferzfg:
I changed it back to keep the `if` statement. This should effectively be an exact match of the old behavior, where we set `GameTimePauseTime` in the setter of `LoadingTimes` here:
https://github.com/LiveSplit/LiveSplit/blob/9cbb008df60b1124f7506762ffbb2462a748b6cd/LiveSplit/LiveSplit.Core/Model/LiveSplitState.cs#L41-L44

Instead, we'll effectively be setting it twice in `SetGameTime` if the Game Time is paused, except the second time it'll be set to a more accurate value.

PR #2287, opened 2022-11-07
Don't parse TimeSpans as floats

### LINKED ISSUE #2286: Start offset imprecision
In 1.8.23 and 1.8.24, start offsets may lose precision. This bug was not present in 1.8.22.

Steps to reproduce:
1. Open the Edit Splits window
2. Enter `5:40.41` in the "Start Timer at:" field
3. Click OK
4. Open the Edit Splits window again
5. The "Start Timer at:" field now contains `5:40.40`

The incorrect offset is also seen in the main window if the layout contains a timer component with 2dp

wooferzfg:
Fix #2286 

There was previously an issue where the number of seconds would be parsed as a floating point number, rather than as an exact number of ticks.

Equivalent PR in livesplit-core: https://github.com/LiveSplit/livesplit-core/pull/578

### DIFF
--- a/LiveSplit/LiveSplit.Core/Model/TimeSpanParser.cs
+++ b/LiveSplit/LiveSplit.Core/Model/TimeSpanParser.cs
@@ -2,6 +2,7 @@
 using System.Linq;
 using System.Globalization;
 using LiveSplit.TimeFormatters;
+using System.Collections.Generic;
 
 namespace LiveSplit.Model
 {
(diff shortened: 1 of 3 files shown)


LiveSplit/LiveSplit.Tests/TimeParseTests/TimeSpanParserTests.cs
```diff
@@ -0,0 +1,35 @@
+﻿using LiveSplit.Model;
+using System;
+using Xunit;
+
+namespace LiveSplit.Tests.TimeParseTests
+{
+    public class TimeSpanParserTests
+    {
+        [Theory]
+        [InlineData("5:40.41", "00:05:40.4100000")]
```
wooferzfg, line 10: This test case was failing with the old code.

LiveSplit/LiveSplit.Core/Model/TimeSpanParser.cs
```diff
@@ -25,12 +26,48 @@ public static TimeSpan Parse(string timeString)
...
+        {
+            { 0, 1L },
+            { 1, 10L },
+            { 2, 100L },
+            { 3, 1000L },
+            { 4, 10000L },
+            { 5, 100000L },
+            { 6, 1000000L },
+            { 7, 10000000L },
+        };
```
wooferzfg, line 71: I wasn't sure about the best way to do `10^x`, since `Math.Pow` takes two `double`s. `BigInteger.Pow` seems to be another option, but it requires importing `System.Numerics`.

CryZe: I really like this approach as it doesn't require any calculations at all. Make it an array though, as that's way faster to access than a dictionary (and it works out in the way the indices are laid out). I'm considering doing the same in lsc if it looks better in benchmarks.

LiveSplit/LiveSplit.Tests/TimeParseTests/TimeSpanParserTests.cs
CryZe: 👍

LiveSplit/LiveSplit.Core/Model/TimeSpanParser.cs
```diff
@@ -25,12 +26,48 @@ public static TimeSpan Parse(string timeString)
...
+
+            return TimeSpan.FromTicks((long)(factor * (secondsTicks + fractionTicks)));
+        }
+
+        private static long ParseFractionAsTicks(string fractionText)
+        {
+            if (fractionText.Length > 7)
+                fractionText = fractionText.Substring(0, 7);
+
+            return long.Parse(fractionText) * powersOfTen[7 - fractionText.Length];
```
CryZe, line 48: I'd recommend using unsigned numbers (ulong iirc) all over the code until the very end where the sign comes in as otherwise you would allow non-sensical stuff like `-01:-23.-45`

LiveSplit/LiveSplit.Core/Model/TimeSpanParser.cs
```diff
@@ -25,12 +26,48 @@ public static TimeSpan Parse(string timeString)
...
+        }
+
+        private static long ParseSecondsAsTicks(string secondsText)
+        {
+            var totalSeconds = secondsText
                 .Split(':')
-                .Select(x => double.Parse(x, NumberStyles.Float, CultureInfo.InvariantCulture))
+                .Select(x => long.Parse(x, NumberStyles.Float, CultureInfo.InvariantCulture))
```
CryZe, line 55: The other parse calls don't take neithr NumberStyles.Float nor CultureInfo. The former seems wrong, maybe, the latter probably makes sense on all.

wooferzfg: Oops, yeah I totally missed this.

PR #2308, opened 2023-01-24
Deep pointer tests

rpgrca:
Add unit tests for some classes and structs.

- AbbreviationExtensions
- AtomicDateTime
- Attempt struct
- IndexedTime
- SegmentHistory
- Segment
- StringExtensions
- Time struct
- LayoutSettings

Only testing files were added/modified.


rpgrca:
Sorry about the pipeline fails, I thought the saved files would have local time values instead of UTC times. Also didn't want to activate fork pipeline just in case, maybe I should next time.

PR #2327, opened 2023-05-07
Add Danger by Design to autosplitters

kclimb:


wooferzfg:
Please use https://github.com/LiveSplit/LiveSplit.AutoSplitters instead

PR #2344, opened 2023-07-20
Fix hard-coded module amount in `ModulesWow64Safe`

### LINKED ISSUE #2343: `LiveSplit.ComponentUtil.ExtensionMethods.ModulesWow64Safe` throws `IndexOutOfRangeException` when `hModules` is not large enough
## Description

https://github.com/LiveSplit/LiveSplit/blob/5a69fb5a41903495ba7e03058b6551c3645776bb/LiveSplit/LiveSplit.Core/ComponentUtil/ProcessExtensions.cs#L59-L65

Today, LiveSplit fetches another process' modules by using `EnumProcessModulesEx`. In its attempt to do so, it assumes a maximum amount of modules of `1024`. This is an incorrect assumption in some rare occasions.

## Fix

Instead, we should call `EnumProcessModulesEx` twice; once to receive the size of the required buffer and another time to actually query the module handles:

```cs
if (!WinAPI.EnumProcessModulesEx(p.Handle, null, 0, out cbNeeded, LIST_MODULES_ALL))
    throw new Win32Exception();

uint numMods = cbNeeded / (uint)IntPtr.Size;
var hModules = new IntPtr[(int)numMods];

if (!WinAPI.EnumProcessModulesEx(p.Handle, hModules, cbNeeded, out _, LIST_MODULES_ALL))
    throw new Win32Exception();
```

just-ero:
Closes #2343 by introducing the fix proposed in the issue.

I wasn't able to build the solution locally unfortunately due to the outdated libraries and references, but I am convinced that everything should work fine regardless.

### DIFF
--- a/LiveSplit/LiveSplit.Core/ComponentUtil/ProcessExtensions.cs
+++ b/LiveSplit/LiveSplit.Core/ComponentUtil/ProcessExtensions.cs
@@ -56,18 +56,19 @@ namespace LiveSplit.ComponentUtil
             const int LIST_MODULES_ALL = 3;
             const int MAX_PATH = 260;
 
-            var hModules = new IntPtr[1024];
-
-            uint cb = (uint)IntPtr.Size*(uint)hModules.Length;
             uint cbNeeded;
-
-            if (!WinAPI.EnumProcessModulesEx(p.Handle, hModules, cb, out cbNeeded, LIST_MODULES_ALL))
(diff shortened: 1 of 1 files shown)


wooferzfg:
@just-ero Are you able to test this? GitHub Actions has build artifacts

just-ero:
The code runs without issues locally.

PR #2365, opened 2023-10-02

shadax1:
added entries for Gensokyo no Nazo, Rosenkreuzstilette and Rosenkreuzstilette Freudenstachel


shadax1:
noticed I'm supposed to do this kind of update [here](https://github.com/LiveSplit/LiveSplit.AutoSplitters), oopsie

Issue #2380, opened 2023-12-19
New game name for "Sonic Adventure (DX)" doesn't seem to integrate properly


slysonic:
Ever since the game name for "Sonic Adventure (DX)" changed to "Sonic Adventure DX: Director's Cut", LiveSplit has been struggling to pull categories and variables (additional info) from speedrun.com. The new game name still appears on the Game Name list, but everything else does not.
![image](https://github.com/LiveSplit/LiveSplit/assets/76589545/62309e81-c3eb-4611-9c6e-d1cb24c7cbb3)
![image](https://github.com/LiveSplit/LiveSplit/assets/76589545/2ca6a0eb-382c-48b6-a9ad-35877e2ffc74)

This issue has persisted since the name changed a few months ago and we're not sure why it happens. It is particularly confusing since later, the game name for "Sonic Adventure (DX) - Category Extensions" got renamed to remove the parenthesis, and LiveSplit has still been able to pull category/variable information as intended from speedrun.com. But the issue still persists with "Sonic Adventure DX: Director's Cut".

wooferzfg:
Same issue as https://github.com/LiveSplit/LiveSplit/issues/1829

wooferzfg:
The issue is that Speedrun.com's search doesn't work properly sometimes. Why does this API call return no results?
https://www.speedrun.com/api/v1/games?name=Sonic%20Adventure%20DX%3A%20Director%27s%20Cut
Yet this one returns results:
https://www.speedrun.com/api/v1/games?name=Sonic%20Adventure%20DX

Issue #2368, opened 2023-11-01
Trackmania Run Categories not showing in Edit Splits window


Rastats:
Good morning,

Recently a new TrackMania game has been added to speedrun.com, it's the original game of the series named **_TrackMania_** (with a capital M) and released in 2003. Problem is that the latest TrackMania game is named **_Trackmania_** (with a lowercase M) and was released in 2020.

Now when I open the "Edit Splits" window on my splits file for the 2020 Trackmania game, the **_Game Name_** selected is still **_Trackmania_** but the **_Run category_** dropdown list doesn't show any categories but only the default ones (**_Any%, Low% and 100%_**).

And now **_Trackmania_** doesn't appear in all the possible registered names for the **_Game Name_** text box and has been replaced with **_Trackmania (2020)_** in the list of registrated games. When I switch the **_Game Name_** to **_Trackmania (2020)_** it says that there is no Autosplitter for this game (which is normal since the Autosplitters database has not been updated with the new game name) and when I have a look at all the "Run Categories", it's also only the 3 default options Any%, Low% and 100%.

However, when I select **_TrackMania (2003)_** as the game name, the **_Run Category_** dropdown does show the two categories (Puzzle and Race) for that specific 2003 game.

Could you please fix this for the Trackmania game from 2020 by keeping Trackmania (2020) as the Game Name in the splits editor, by associating the run categories from the [https://speedrun.com/tm](speedrun.com/tm) leaderboards to that game name and by editing the autosplitter database with the updated game name ?

Thanks in advance,

Rastats

LeddaZ:
Same issue for Super Mario Bros. Wonder

wooferzfg:
@Rastats Can you fix this by setting the game name in your splits file to "Trackmania (2020)" and adding that as a game name to the autosplitter XML (https://github.com/LiveSplit/LiveSplit.AutoSplitters/blob/master/LiveSplit.AutoSplitters.xml#L8085)?

wooferzfg:
@LeddaZ What issue are you having with Super Mario Bros. Wonder? That game only has one entry on Speedrun.com with that name, and no autosplitter

Rastats:
I updated the autosplitter XML, I still only have the Any%, 100% and Low% options in the Splits editor for both "Trackmania" and "Trackmania (2020)" as the game name.

LeddaZ:
[...]

PR #2387, opened 2024-01-07
Get speedrun.com game categories using their game ID, not their game name

### LINKED ISSUE #1829: Some Speedrun.com games can't be found
We rely on Speedrun.com's search, which doesn't always return the correct game, even if you have an exact match. We should be using the game ID instead.

Refragg:
This patch changes the way the categories for a game on speedrun.com are retrieved. Previously, it would try searching using the name which works in most cases but there are some edge cases that are mishandled by the speedrun.com API. Now, the categories will be retrieved using the game ID which will be previously cached by the CompositeGameList inside a dictionary. This change only adds a field and 2 methods to the CompositeGameList class which shouldn't break applications relying on the previous behavior.

This fixes #1829, #2368 and #2380

### DIFF
--- a/LiveSplit/LiveSplit.Core/Model/RunMetadata.cs
+++ b/LiveSplit/LiveSplit.Core/Model/RunMetadata.cs
@@ -178,7 +178,8 @@ namespace LiveSplit.Model
                         {
                             try
                             {
-                                var game = SpeedrunCom.Client.Games.SearchGameExact(oldGameName, new GameEmbeds(embedRegions: true, embedPlatforms: true));
+                                var gameName = Web.CompositeGameList.Instance.GetGameID(oldGameName);
+                                var game = SpeedrunCom.Client.Games.GetGame(gameName, new GameEmbeds(embedRegions: true, embedPlatforms: true));
                                 gameLoaded = true;
                                 if (game != null)
                                     GameAvailable = true;
--- a/LiveSplit/LiveSplit.Core/Model/RunMetadata.cs
+++ b/LiveSplit/LiveSplit.Core/Model/RunMetadata.cs
@@ -178,8 +178,8 @@ namespace LiveSplit.Model
                         {
                             try
                             {
-                                var gameName = Web.CompositeGameList.Instance.GetGameID(oldGameName);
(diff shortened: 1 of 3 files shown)


LiveSplit/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -178,7 +178,8 @@ public void Refresh()
                         {
                             try
                             {
-                                var game = SpeedrunCom.Client.Games.SearchGameExact(oldGameName, new GameEmbeds(embedRegions: true, embedPlatforms: true));
+                                var gameName = Web.CompositeGameList.Instance.GetGameID(oldGameName);
+                                var game = SpeedrunCom.Client.Games.GetGame(gameName, new GameEmbeds(embedRegions: true, embedPlatforms: true));
```
wooferzfg, line 182: Can you call this variable `gameId` instead of `gameName`?

LiveSplit/LiveSplit.Core/Web/CompositeGameList.cs
```diff
@@ -45,5 +49,38 @@ public IEnumerable<string> GetGameNames()
 
             return gameNames;
         }
+        
+        public IList<string> GetGameNamesAndCacheIDs()
```
wooferzfg, line 53: Why add a separate method instead of updating `GetGameNames` to cache the IDs? Can we remove `GetGameNames` after this change?

wooferzfg:
Thanks for this PR, this will be really helpful.

LiveSplit/LiveSplit.Core/Web/CompositeGameList.cs
```diff
@@ -13,9 +13,13 @@ public class CompositeGameList
 
         protected IList<string> gameNames;
 
+        protected IDictionary<string, string> gameIDs;
+
         protected CompositeGameList()
         { }
 
+        public string GetGameID(string gameName) => gameIDs[gameName];
```
wooferzfg, line 21: The issue with this approach is that we require all the game names to be loaded before we can get a game ID. This means that if you open a new set of splits but don't open the run editor, then none of the metadata is accessible (e.g. the World Record component doesn't work).

PR #2389, opened 2024-01-12
Update copyright years in README and LICENSE

developerrowan:
README's end date still had 2023 (not a huge deal), LICENSE contained an end date of 2020 (not really a huge deal either).

Purely a cosmetic change!

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2013-2024 Christopher Serr and Sergey Papushin
+Copyright (c) 2013 Christopher Serr and Sergey Papushin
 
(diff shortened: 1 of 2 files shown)


CryZe:
I'm wondering if we even want to specify anything like that in the first place? I don't think it has any meaningful legal impact or anything and is annoying enough to keep up to date.

developerrowan:
Given the license is MIT, the impact of dates is none whatsoever. I would agree it might be nicer to not have to keep after it.

wooferzfg:
Could we just write `2013-`?

CryZe:
Just `2013` should be enough. That's how most projects seem to do it.

developerrowan:
Change made! Agree with this decision. Sorry for the annoying PR! 😄

PR #2395, opened 2024-02-02
Splits with true zero delta is now considered ahead

OCircles:
Good visual feedback for IGT runs

### DIFF
--- a/LiveSplit/LiveSplit.Core/Model/LiveSplitStateHelper.cs
+++ b/LiveSplit/LiveSplit.Core/Model/LiveSplitStateHelper.cs
@@ -154,7 +154,7 @@ namespace LiveSplit.Model
 
             if (timeDifference != null)
             {
-                if (timeDifference < TimeSpan.Zero)
+                if (timeDifference <= TimeSpan.Zero)
                 {
                     splitColor = state.LayoutSettings.AheadGainingTimeColor;
                     var lastDelta = GetLastDelta(state, splitNumber - 1, comparison, method);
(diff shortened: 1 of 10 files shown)


OCircles:
Ah, that's a lot of failed "0.0" checks. You could edit to this maybe?

`minusString = (time == TimeSpan.Zero ? "" : TimeFormatConstants.MINUS);`

Looks like this

Ans4EE2

I don't mind the lack of minus as much as I thought I would.

LiveSplit/LiveSplit.Core/Model/LiveSplitStateHelper.cs
```diff
@@ -154,12 +154,14 @@ public static class LiveSplitStateHelper
...
             {
-                if (timeDifference < TimeSpan.Zero)
+                if (timeDifference <= TimeSpan.Zero)
                 {
                     splitColor = state.LayoutSettings.AheadGainingTimeColor;
                     var lastDelta = GetLastDelta(state, splitNumber - 1, comparison, method);
                     if (showSegmentDeltas && splitNumber > 0 && lastDelta != null && timeDifference > lastDelta)
                         splitColor = state.LayoutSettings.AheadLosingTimeColor;
+                    if (timeDifference == TimeSpan.Zero)
+                        splitColor = state.LayoutSettings.TextColor;
```
wooferzfg, line 164: This should just be a separate `if` instead of being inside `if (timeDifference <= TimeSpan.Zero)`

LiveSplit/LiveSplit.Core/TimeFormatters/GeneralTimeFormatter.cs
```diff
@@ -61,9 +61,9 @@ public string Format(TimeSpan? timeNullable)
             TimeSpan time = timeNullable.Value;
 
             string minusString;
-            if (time < TimeSpan.Zero)
+            if (time <= TimeSpan.Zero)
             {
-                minusString = TimeFormatConstants.MINUS;
+                minusString = (time == TimeSpan.Zero ? "" : TimeFormatConstants.MINUS);
```
wooferzfg, line 66: Same thing here, make this a separate check instead of jamming it into the existing code

wooferzfg:
livesplit-core change: https://github.com/LiveSplit/livesplit-core/pull/768

wooferzfg:
Probably best not to add a new setting unless that's the plan for livesplit-core as well.

CryZe:
Yeah we discussed that the additional color is overkill because the chance of this happening is incredibly rare. We already use the text color for a skipped delta, so using the same color would make sense. No idea why the commit got pushed anyway.

PR #2398, opened 2024-02-07
Splits with true zero delta get ahead color, no sign

AlexKnauth:
An updated version of #2395 by @OCircles, but with more tests for exact `0` vs `+0`, and with separate `if` branches for exact `0`.


LiveSplit/LiveSplit.Core/Model/LiveSplitStateHelper.cs
```diff
@@ -154,7 +154,11 @@ public static class LiveSplitStateHelper
 
             if (timeDifference != null)
             {
-                if (timeDifference < TimeSpan.Zero)
+                if (timeDifference == TimeSpan.Zero)
+                {
+                    splitColor = state.LayoutSettings.TextColor;
```
wooferzfg, line 159: Here's all the uses of `GetSplitColor`: https://github.com/search?q=org%3ALiveSplit%20GetSplitColor&type=code

Some components apply an override to the times color if this method returns `null`, so we should probably return `null` here.
However, the previous segment component currently does not handle this method returning `null`, so it needs to be updated to handle that (and just use `state.LayoutSettings.TextColor` in that case).

wooferzfg: I'm also not sure if this change will break any custom components

AlexKnauth: Is that null override desirable for true zero deltas?

Before this change, `GetSplitColor` returned either the behind-loosing-time color or the behind-gaining-time color, so existing components won't be expecting a null to override in this situation.

wooferzfg: Returning the text color in the layout settings isn't correct because some components have an override for the time color

wooferzfg: It might be simpler to just return the ahead color when the delta is exactly 0 (so just change the condition for the ahead color to `<= TimeSpan.Zero`), instead of a separate color for a 0.0 delta

wooferzfg: The image from the other PR is a good example of what it would look like:

Ans4EE2

AlexKnauth: Ah, I see in the timer component, this code:
```C#
TimerColor = LiveSplitStateHelper.GetSplitColor(state, state.CurrentTime[timingMethod] - state.CurrentSplit.Comparisons[state.CurrentComparison][timingMethod],
    state.CurrentSplitIndex, true, false, state.CurrentComparison, timingMethod)
    ?? state.LayoutSettings.AheadGainingTimeColor;
```
Overrides null to the ahead-gaining-time color, which is desirable because that light green looks nicer on the timer than the neutral white that I saw there while testing.

AlexKnauth: After `GetSplitColor exact 0 null`, the timer component is back to showing the light green ahead-gaining-time color to signal that the timer has started, even when the delta is on exact 0.

wooferzfg: > It might be simpler to just return the ahead color when the delta is exactly 0

I think this should give the same behavior in the timer component?

AlexKnauth: The timer component should show the ahead color either way. But in the deltas in the splits component, it seems better to me to show the neutral text color. That's what the livesplit-core version of this change already did.

wooferzfg: I'd like to avoid breaking custom components with this change, though. livesplit-core doesn't have custom components, so it can change the implementation freely

wooferzfg: I'm also concerned with some people not liking the new behavior. The change to the ahead color seems less controversial, since it doesn't introduce a new color

wooferzfg:
Thanks for putting this change together and being flexible

PR #2399, opened 2024-02-10
Make the command server use and accept the standard C# TimeSpan format.

TheTedder:
Adds a new `ITimeFormatter` that formats `TimeSpan`s according to C#'s "c" format specifier. Makes the command server use this formatter for outputting times. Makes the time parser more liberal in what it can parse by adding support for days, specified by an integer followed by a period before the rest of the formatted time string.

### DIFF
--- a/LiveSplit/LiveSplit.Core/Model/TimeSpanParser.cs
+++ b/LiveSplit/LiveSplit.Core/Model/TimeSpanParser.cs
@@ -8,6 +8,9 @@ namespace LiveSplit.Model
 {
     public static class TimeSpanParser
     {
+        private static readonly char[] separators = { ':' };
+        private static readonly char[] dot = { '.' };
+
         public static TimeSpan? ParseNullable(string timeString)
         {
             if (string.IsNullOrEmpty(timeString))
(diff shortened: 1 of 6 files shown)


LiveSplit/LiveSplit.Core/Model/TimeSpanParser.cs
```diff
@@ -26,14 +26,20 @@ public static TimeSpan Parse(string timeString)
                 timeString = timeString.Substring(1);
             }
 
-            var splitTimeString = timeString.Split(new char[] {'.'}, 2);
-            var secondsText = splitTimeString[0];
+            var splitTimeString = timeString.Split(new char[] {'.'}, 3);
+            var secondsText = splitTimeString[splitTimeString.Length - 2];
```
wooferzfg, line 30: What if `splitTimeString.Length` is less than 2?

wooferzfg: Could you add tests for this class if you're making changes to it, since it's used in places besides `CommandServer`?

LiveSplit/LiveSplit.Core/TimeFormatters/PreciseTimeFormatter.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace LiveSplit.TimeFormatters
+{
+    /// <summary>
+    /// Formats <see cref="TimeSpan"/>s according to C#'s built-in "c" format, using a hyphen for null TimeSpans.
+    /// </summary>
+    public class PreciseTimeFormatter : ITimeFormatter
```
wooferzfg, line 12: Maybe something like `CTimeFormatter` would be more descriptive

CryZe: I recommended this name because the intention is not to match C# (which is a side effect) but to have as many digits as possible to not lose out on precision.

LiveSplit/LiveSplit.Core/Model/TimeSpanParser.cs
```diff
@@ -19,55 +22,62 @@ public static TimeSpan Parse(string timeString)
...
+
+                    if (daysDotHours.Length == 2)
+                    {
+                        ticks += long.Parse(daysDotHours[0], CultureInfo.InvariantCulture) * TimeSpan.TicksPerDay;
+                    }
+
+                    ticks += long.Parse(daysDotHours[daysDotHours.Length - 1], CultureInfo.InvariantCulture) * TimeSpan.TicksPerHour;
+                    goto case 2;
+                case 2:
+                    ticks += long.Parse(sections[sections.Length - 2], CultureInfo.InvariantCulture) * TimeSpan.TicksPerMinute;
```
wooferzfg, line 48: This code is hard to read, maybe split it up into separate methods?

Does the livesplit-core version of this code have the same behavior? It looks a lot s
cleaner https://github.com/LiveSplit/livesplit-core/blob/d8ba361d97e975c1ba5302e77c9833dc95cd9855/src/timing/time_span.rs#L110

TheTedder: I'm pretty sure the lscore version doesn't have to worry about days at all. That's the main source of complexity. Should I change it to use floating point math instead?

CryZe: Yeah lsc handles it here: https://github.com/LiveSplit/livesplit-core/blob/d8ba361d97e975c1ba5302e77c9833dc95cd9855/src/run/parser/livesplit.rs#L162
(i.e. the lss parser)

No, don't use floating point.

wooferzfg: Ah okay, makes sense

I think avoiding floating point math would be ideal. But at least splitting this up into smaller functions and avoiding a `goto` would be good for readability

PR #2408, opened 2024-03-01
Patch 1

shmumbler:


SpeedyFolf:
You should close this pull request and open a new one at https://github.com/LiveSplit/LiveSplit.AutoSplitters instead. Also you shouldn't name your pull request "Patch 1" or "Update LiveSplit.AutoSplitters.xml", you should name it something like "Add load-remover for NecroVisioN games" instead.

PR #2414, opened 2024-03-28
Improve Server Error Handling

TheTedder:
This PR fixes a few bugs such as threading issues and greatly improves error handling of the server in cases of bad command syntax. There are still a few known issues related to connection handling so there will be at least one more PR—probably two.


wooferzfg:
This was tested with the apps that use the server?

TheTedder:
> This was tested with the apps that use the server?

This PR in particular doesn't make them any more broken. I can tell you that. It's mostly about fixing internal issues.

PR #2425, opened 2024-03-31
Convert `NAudio` to NuGet package

just-ero:
### Description

Drops the physical `NAudio` dependency in `Libs` in favor of a NuGet package. This change affects `LiveSplit.Sound`.


just-ero:
Failure is the missing `NAudio` dependency for building `LiveSplit.Sound`.

PR #2428, opened 2024-03-31
Remove `CLRHost` binaries

just-ero:
### Description

Removes the `CLRHost` binaries, which aren't used anywhere.


wooferzfg:
This was originally added for the OBS plugin.

PR #2432, opened 2024-04-01
Migrate to `PackageReference`

just-ero:
### Description

Removes `packages.config` in favor of `PackageReference`s.


LiveSplit/LiveSplit.View/LiveSplit.View.csproj
```diff
@@ -326,7 +325,11 @@
       <EmbedInteropTypes>True</EmbedInteropTypes>
     </COMReference>
   </ItemGroup>
-  <ItemGroup />
+  <ItemGroup>
+    <PackageReference Include="WindowsAPICodePack">
+      <Version>1.1.0</Version>
+    </PackageReference>
```
just-ero, line 331: Need to figure out how to exclude certain files from output. The output currently contains
```
Microsoft.WindowsAPICodePack.dll
Microsoft.WindowsAPICodePack.ExtendedLinguisticServices.dll
Microsoft.WindowsAPICodePack.Sensors.dll
Microsoft.WindowsAPICodePack.Shell.dll
Microsoft.WindowsAPICodePack.ShellExtensions.dll
```
The goal is to only include
```
Microsoft.WindowsAPICodePack.dll
Microsoft.WindowsAPICodePack.Shell.dll
```

wooferzfg: Not sure if this would work in this case, but we currently remove some extra files from the output in the CI (not ideal even if it works, though) https://github.com/LiveSplit/LiveSplit/blob/82cb84863678d5296075d77b91d845fb80616eaf/.github/workflows/ci.yml#L54-L61

Regardless, even if we can't figure out how to replicate the old output behavior, a few extra files isn't that bad

wooferzfg: Just tested deleting `Microsoft.WindowsAPICodePack.ExtendedLinguisticServices.dll`, `Microsoft.WindowsAPICodePack.Sensors.dll`, `Microsoft.WindowsAPICodePack.ShellExtensions.dll`, and it does work

just-ero: It appears that `PackageReference` is not capable of excluding any output content. We will have to delete the files after the fact.

just-ero:
I think the files might be in `Components`? Have you checked for that?

LiveSplit/LiveSplit.View/LiveSplit.View.csproj
wooferzfg: I pushed a change to the CI workflow that removes the extra files

wooferzfg:
They were in the main folder because they're imported by the main LiveSplit code rather than one of the components

PR #2441, opened 2024-04-07
Update `git` config files

just-ero:
### Description

- Removed `.gitattributes`.
- Simplified `.gitignore` and applied exclusions.


.gitignore
```diff
@@ -1,251 +1,22 @@
...
-LiveSplit/LiveSplit.Core/Updates/GitInfo/version.txt
-LiveSplit/LiveSplit.Core/Updates/GitInfo/revision.txt
-.vs
+# LiveSplit
+GitInfo/
 
-LiveSplit/LiveSplit.Core/x64/target
-LiveSplit/LiveSplit.Core/x86/target
+x64/target/
+x86/target
```
wooferzfg, line 22: Why is x86 different from x64 here?

.gitignore
```diff
@@ -1,251 +1,22 @@
...
-FakesAssemblies/
-
-# =========================
-# Operating System Files
-# =========================
-
-# OSX
-# =========================
-
-.DS_Store
```
wooferzfg, line 203: Please keep this because someone might edit code on Mac even if they don't built it there

.gitignore
wooferzfg: I think the x64 isn't working. `LiveSplit\LiveSplit.Core\x64\target\release\.fingerprint` isn't ignored for me

PR #2442, opened 2024-04-08
Switch to a cleaned up SVG icon

CryZe:
The LiveSplit icon was always just manually put together in GIMP and all the circles were placed by hand. I recently wrote a small script that mathematically correctly places all elements of the icon and this time emits it as the smallest possible vectorized image. This means the LiveSplit icon is now also available at any resolution.

The code can be found here: https://github.com/CryZe/livesplit-icon-generator


wooferzfg:
There's also these files:
https://github.com/LiveSplit/LiveSplit/blob/master/LiveSplit/LiveSplit.View/Resources/Icon.png
https://github.com/LiveSplit/LiveSplit/blob/master/LiveSplit/LiveSplit.View/Resources/Icon.ico

But I'm not sure if there's some issue with replacing these with an SVG

PR #2443, opened 2024-04-08
Remove unused files in Resources

wooferzfg:
These files aren't referenced by any of the code. `LiveSplit.View/Resources` has copies of some of these files


CryZe:
The xcfs are specifically committed so we can make changes to stuff. Icon_mip.xcf for example is to generate the Icon.ico.

PR #2444, opened 2024-04-08
Update SRC API key link to not include their name

YummyBacon5:
I've updated the SRC API key link to not include the username. Since this link redirects directly to the user's page without them having to type in their name.


wooferzfg:
Thanks, I didn't realize there was still a way to get to this page since they broke the `speedrun.com/me/settings/api` link. I updated the process to just open to the correct URL instead of having "go to" in the text.

PR #2451, opened 2024-04-14
Revert PlatformTarget change

wooferzfg:
In https://github.com/LiveSplit/LiveSplit/commit/4b0ccc3085b559b7172d28eaede31dc0ef064c56, we started building LiveSplit.exe to support 64 bit only. I believe the previous behavior was "Any CPU" based on https://stackoverflow.com/a/30352426

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,6 @@ jobs:
           dotnet build
           -v m
           -c Release
-          -p:PlatformTarget=x64
           -p:GenerateDocumentationFile=false
           -p:DebugType=None
           .\LiveSplit.sln
@@ -45,7 +44,6 @@ jobs:
           dotnet build
           -v m
           -c UpdateManagerExe
-          -p:PlatformTarget=x64
           -p:GenerateDocumentationFile=false
           -p:DebugType=None
           .\src\UpdateManager\UpdateManager.csproj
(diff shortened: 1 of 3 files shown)


just-ero:
AnyCPU is the default, it doesn't need to be explicitly set.

The `PlatformTarget` was set due to something about webview. I believe not setting it will produce more files in `Components/runtimes`.

wooferzfg:
> AnyCPU is the default, it doesn't need to be explicitly set.

I set it explicitly to make it clearer what's being used. Also, the default when I edit the project files in Visual Studio is `x86`

> The PlatformTarget was set due to something about webview. I believe not setting it will produce more files in Components/runtimes.

I think those extra files are to get it running on different types of processors. We do the same thing for `asr_capi.dll` and `livesplit_core.dll` already.

just-ero:
And `x86` *will* run on *any* CPU. You cannot actually build an app that's somehow both x64 and x86 at the same time.

wooferzfg:
`x86` and `AnyCPU` aren't the same thing https://stackoverflow.com/a/516740

just-ero:
The post you linked is talking about *assemblies* being loaded into a *process*. The process (`LiveSplit.exe`) must still be x86 or x64, making the decision for the target architecture void.

wooferzfg:
> Considering most companies don't even offer x86 versions anymore, I don't believe we should either.

Makes sense. I think it's worth discussing in the Discord before making the change. Also, if we do that, we should remove the x86 DLLs for ASR and livesplit-core.

just-ero:
cc @CryZe

PR #2452, opened 2024-04-14
Remove GenerateDocumentationFile override from command line args

wooferzfg:
Simplifying the CI args

### DIFF
--- a/src/LiveSplit.Core/LiveSplit.Core.csproj
+++ b/src/LiveSplit.Core/LiveSplit.Core.csproj
@@ -7,10 +7,13 @@
     <UseWindowsForms>true</UseWindowsForms>
 
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <GenerateResourceUsePreserializedResources>true</GenerateResourceUsePreserializedResources>
     <Nullable>disable</Nullable>
     <NoWarn>1587;1591;1701;1702</NoWarn>
+  </PropertyGroup>
+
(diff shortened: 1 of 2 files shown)


src/LiveSplit.Core/LiveSplit.Core.csproj
```diff
@@ -7,12 +7,15 @@
     <UseWindowsForms>true</UseWindowsForms>
 
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
just-ero, line 10: I do not understand this change.

wooferzfg: I updated the configuration so that the documentation is only generated in the Debug configuration and not for Release. That matches what the CI was already doing

just-ero: That's my bad, perhaps. The user obviously wants the documentation comments while they use libraries like LiveSplit.Core in their components.

wooferzfg: We've never included any XMLs in the release build file, though

just-ero: Hmm. Then can we not just turn `GenerateDocumentationFile` off? Pretty sure that would remove all the `NoWarn` elements as well.

wooferzfg: https://github.com/LiveSplit/LiveSplit/commit/947f499fb1ee738bbbf0e1e3fbfd3bf5fbeee8fa

Issue #2454, opened 2024-04-17
Prevent user from accidentally overwriting a splits file with a layout file


flemgus:
every single time i save a livesplit thing it always tells me that when i load it up, this is the most recent version and i re-installed it and tried re making it 3 times
![image_2024-04-17_205316929](https://github.com/LiveSplit/LiveSplit/assets/167311381/cf0abc9f-9c55-4dbb-9b13-ed8073b5afb5)

CryZe:
Please attach the splits file as well, so we can look into it.

TheTaraStark:
I've also had this issue since switching from LSO to LS, I think it's something to do with the image files used for icons because clearing the icons and uploading different images fixed the issue for me. Sadly, git won't let me attach an .lss file but I'll try to get something in the Discord

CryZe:
So we just fixed the bug @TheTaraStark encountered, but I'm not convinced that this fixes the original bug. So I'd still like a splits file for that.

just-ero:
We should expand #2371 to include all image formats that LiveSplit One supports.

flemgus:
> Please attach the splits file as well, so we can look into it.

yes but it said that github doesn't support that file type, is there any other way i can send it?

flemgus:
> Please attach the splits file as well, so we can look into it.

[Subnautica - Any% - Copy.zip](https://github.com/LiveSplit/LiveSplit/files/15022117/Subnautica.-.Any.-.Copy.zip)
 here's a compressed version

CryZe:
That's a layout file. It seems like you accidentally overwrote your splits with the layout.

CryZe:
@wooferzfg Maybe we should error out in case the file extension is mixed up like that before overwriting any files.

flemgus:
> That's a layout file. It seems like you accidentally overwrote your splits with the layout.

ah yes how do i fix that or do it the proper way again?

flemgus:
> @wooferzfg Maybe we should error out in case the file extension is mixed up like that before overwriting any files.

yes oaky i re did them and it worked thank you idk what went wrong beofre but it's fixed now

PR #2455, opened 2024-04-17
Skip incompatible image formats

CryZe:
When parsing the splits we may encounter splits from LiveSplit One where images may be using newer image formats such as WEBP, AVIF, and JPEG XL. Instead of rejecting the splits files entirely, we simply skip those images instead. A long term solution could be to try converting those images into PNG.

cc https://github.com/LiveSplit/LiveSplit/issues/2454 (probably doesn't fix the main issue but the one in the comment).

### DIFF
--- a/src/LiveSplit.Core/Model/RunFactories/StandardFormatsRunFactory.cs
+++ b/src/LiveSplit.Core/Model/RunFactories/StandardFormatsRunFactory.cs
@@ -1,4 +1,5 @@
 ﻿using LiveSplit.Model.Comparisons;
+using LiveSplit.Options;
 using LiveSplit.UI;
 using System;
 using System.Drawing;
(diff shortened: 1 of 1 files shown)


src/LiveSplit.Core/Model/RunFactories/StandardFormatsRunFactory.cs
```diff
@@ -78,7 +79,19 @@ static Image ParseImage(IntPtr imagePtr, ulong length)
...
+            {
+                return Image.FromStream(stream);
+            }
+            catch (Exception ex)
+            {
+                // .NET Framework's Image doesn't support newer image formats
+                // such as AVIF, WEBP, and JPEG XL. These may be used in
+                // LiveSplit One. So we need to fall back to an empty image
+                // instead of erroring out.
+                Log.Error(ex);
```
wooferzfg, line 92: Do we want to dispose `stream` here, since we're not using the image?

CryZe: Oh yeah that would make sense.

just-ero: You would want to dispose it anyway, no? Is the steam needed after creating the image?

wooferzfg (approved):

PR #2456, opened 2024-04-18
Server: added command addloadingtimes

king-dedede1:
Added a server command that adds loading time, for when you need to remove a specific amount of time from game time

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ Commands that generate no response:
 - starttimer
 - setgametime TIME
 - setloadingtimes TIME
+- addloadingtimes TIME
 - pausegametime
 - unpausegametime
 - alwayspausegametime
--- a/src/LiveSplit.Core/Server/CommandServer.cs
+++ b/src/LiveSplit.Core/Server/CommandServer.cs
@@ -235,6 +235,20 @@ namespace LiveSplit.Server
                         }
                         break;
                     }
+                case "addloadingtimes":
+                    {
+                        try
+                        {
+                            var time = ParseTime(args[1]);
+                            State.LoadingTimes = State.LoadingTimes.Add(time ?? TimeSpan.Zero);
+                        }
+                        catch (Exception e)
+                        {
+                            Log.Error(e);
+                            Log.Error($"[Server] Failed to parse time while adding loading times: {args[1]}");
+                        }
+                        break;
+                    }
                 case "pausegametime":
                     {
                         State.IsGameTimePaused = true;


king-dedede1:
If you needed to subtract 1 second from game time, you'd have to call `getloadingtimes` , parse the string into a time, add 1 second, convert the time back into a string, and then call `setloadingtimes`. A command like `addloadingtimes` would consolidate the two calls into one, which would save time, and make the client implementation a lot simpler because you wouldn't have to do the conversion from string to time span, which could be prone to errors

just-ero:
I don't necessarily agree with combining both game time and loading times; it's generally not something that makes sense.

You would either keep track of the current game time and use `setgametime`, or you would keep track of the loading times and use `setloadingtimes`.

No authority on what gets added or anything, just feel like it shouldn't be LiveSplit's duty to handle any other case.

wooferzfg:
> If you needed to subtract 1 second from game time, you'd have to call getloadingtimes , parse the string into a time

Shouldn't your client be tracking the loading times internally?

king-dedede1:
I made an autosplitter for a game, and there's a timing rule that says when you get a specific load screen that's slightly longer than the others, you remove 1 second from your time.  I wasn't sure how to implement that in an autosplitter, because as far as I can tell the only way to remove game time in ASL is with the `isLoading` action or the `gameTime` action, and neither of those were what I was looking for. I tried using a `System.Windows.Forms.Timer` to pause the game time with `isLoading` then 1 second later unpause, something like this: 

```
    vars.timer = new System.Windows.Forms.Timer();
    vars.timer.Enabled = false;

    vars.timer.Interval = 1000;
    vars.timer.Tick += new EventHandler((sender, e) => {
        vars.shouldStopTimer = false;
        vars.timer.Enabled = false;
    });
```
However it isn't accurate enough, it seems to remove a slightly different amount of time every time, usually a little more than one second. Eventually I realized you could add loading time directly with the `timer` variable:
```
timer.SetGameTime(timer.CurrentTime.GameTime.Value.Subtract(TimeSpan.FromSeconds(1)));
```
and that seemed to work how I wanted it to, so it's the solution I went with. Now, I'm rewriting the autosplitter to use an external program to read the game's memory and use the server to control livesplit, but there isn't a way to directly add loading time like how I did before. 

There could also be some cases where you need to combine pausing/unpausing game time with adding loading time, and in that case the client wouldn't know what the current loading time is so it couldn't keep track of it itself, so I think a command like `addloadingtimes` would be a better solution.

PR #2472, opened 2024-04-20
Update workflows

cloudcloudskies:
Just updated the versions, the functionality is same.


just-ero:
This updates absolutely nothing. The tags you inserted are pointed at by the currently used tags already. There is no need for this change.

Issue #2448, opened 2024-04-13
Font size units are not aligned


Kuhaku9893:
When performing new Font(), the unit of font size may be pixel or piont.
In the case of point, some may explicitly specify the unit while others may not (i.e., default to point).
https://github.com/search?q=repo%3ALiveSplit%2FLiveSplit%20%22new%20Font(%22&type=code

Different font size units seem to cause the following behavior.

For example, in the Text component,
- The first time a CustomFontDialog is opened, if the user reduces the size from 13 to 11, the actual text displayed will be larger.
- The first time a CustomFontDialog is opened, if the user selects a different size and then resets it back to the original size, the text displayed on the timer will be larger than the original size.
- The default font size unit is pixel, but if you change any setting in CustomFontDialog, the font size unit will be point.

Kuhaku9893:
Default values for layout settings are also not consistent.
For example, the font sizes for Times and Text are,
- The default setting when LiveSplit is launched is 13 pixels.
-DefaultLayout.lsl uses 12 points.

StandardLayoutSettingsFactory.cs
https://github.com/LiveSplit/LiveSplit/blob/9b0fcebcb3a75fa452a9cf7c534ec91adef69091/src/LiveSplit.Core/Options/SettingsFactories/StandardLayoutSettingsFactory.cs#L28

DefaultLayout.lsl
https://github.com/LiveSplit/LiveSplit/blob/b3dd60b8054e44c67f1a939f5ae48ea4a2244d0a/src/LiveSplit.View/Resources/DefaultLayout.lsl#L25

PR #2476, opened 2024-04-21
Fix font size units are not aligned

Kuhaku9893:
Related issue: https://github.com/LiveSplit/LiveSplit/issues/2448

- Changed font size unit to pixel after editing in CustomFontDialog.
- Changed font size unit of DefaultLayout.lsl to pixel (e.g. Text font size 12 points → 16 pixels).


wooferzfg:
The timer font size is a tiny bit different now. Is it possible to make it the same as before?
LiveSplit_Jp3AQbBc65
LiveSplit_QVNENHm14e

wooferzfg:
If I open [the LiveSplit build based on your changes](https://github.com/LiveSplit/LiveSplit/actions/runs/8773960982/artifacts/1433400209), go to "Edit Layout" -> "Layout Settings" -> "Timer Font", it shows this:
LiveSplit_PfoI0dw8Js

However, if I instead go to "Open Layout" -> "Default" and then "Edit Layout" -> "Layout Settings" -> "Timer Font", it shows this:
LiveSplit_iaBzakDuIc

I have the same issue with the Title component (the font size is different in the default layout vs. if I add the Title component before opening the default layout).

You might need to update the defaults here? https://github.com/LiveSplit/LiveSplit/blob/0b8a66d38f08bd38731d8756f8fdf98986e3363b/src/LiveSplit.Core/Options/SettingsFactories/StandardLayoutSettingsFactory.cs#L27-L29

just-ero:
Which one is the font size unit used for the font dialog?

wooferzfg:
> Which one is the font size unit used for the font dialog?

I think it depends on which font size unit the font has before you open the dialog?

It seems like this change only affects the sample text in the dialog, and not the font that gets passed to LiveSplit: https://github.com/LiveSplit/LiveSplit/pull/2476/files#diff-bd68a407dcc17911faafc1b285b2a55e5e997f7150b18e575ee8d181a2fd7c6bL169-R169

wooferzfg:
@Kuhaku9893 Have you tested your changes in this PR? I'm not sure that it fixes the bugs you mentioned in #2448 

Also, I pushed a change to have `LiveSplit.sln` include the `CustomFontDialog` project https://github.com/LiveSplit/LiveSplit/commit/0b8a66d38f08bd38731d8756f8fdf98986e3363b
Before that change, you needed to update `CustomFontDialog.dll` for the changes to take effect.

wooferzfg:
> Which one is the font size unit used for the font dialog?

Before this PR, it switched to Point after you made a change in the dialog. After this PR, it should always use Pixel after making changes in the dialog.

However, there's some other code that's causing strange behavior. For example: https://github.com/LiveSplit/LiveSplit/blob/56f61c457630ff7283e9bb307961a4d49c7e1c3c/src/LiveSplit.View/View/LayoutSettingsControl.cs#L140-L141

src/LiveSplit.View/Resources/DefaultLayout.lsl
```diff
@@ -22,9 +22,9 @@
...
+    <TimesFont><![CDATA[AAEAAAD/////AQAAAAAAAAAMAgAAAFFTeXN0ZW0uRHJhd2luZywgVmVyc2lvbj00LjAuMC4wLCBDdWx0dXJlPW5ldXRyYWwsIFB1YmxpY0tleVRva2VuPWIwM2Y1ZjdmMTFkNTBhM2EFAQAAABNTeXN0ZW0uRHJhd2luZy5Gb250BAAAAAROYW1lBFNpemUFU3R5bGUEVW5pdAEABAQLGFN5c3RlbS5EcmF3aW5nLkZvbnRTdHlsZQIAAAAbU3lzdGVtLkRyYXdpbmcuR3JhcGhpY3NVbml0AgAAAAIAAAAGAwAAAAhTZWdvZSBVSQAAgEEF/P///xhTeXN0ZW0uRHJhd2luZy5Gb250U3R5bGUBAAAAB3ZhbHVlX18ACAIAAAABAAAABfv///8bU3lzdGVtLkRyYXdpbmcuR3JhcGhpY3NVbml0AQAAAAd2YWx1ZV9fAAgCAAAAAgAAAAs=]]></TimesFont>
+    <TimerFont><![CDATA[AAEAAAD/////AQAAAAAAAAAMAgAAAFFTeXN0ZW0uRHJhd2luZywgVmVyc2lvbj00LjAuMC4wLCBDdWx0dXJlPW5ldXRyYWwsIFB1YmxpY0tleVRva2VuPWIwM2Y1ZjdmMTFkNTBhM2EFAQAAABNTeXN0ZW0uRHJhd2luZy5Gb250BAAAAAROYW1lBFNpemUFU3R5bGUEVW5pdAEABAQLGFN5c3RlbS5EcmF3aW5nLkZvbnRTdHlsZQIAAAAbU3lzdGVtLkRyYXdpbmcuR3JhcGhpY3NVbml0AgAAAAIAAAAGAwAAAA5DZW50dXJ5IEdvdGhpY1VVaUIF/P///xhTeXN0ZW0uRHJhd2luZy5Gb250U3R5bGUBAAAAB3ZhbHVlX18ACAIAAAABAAAABfv///8bU3lzdGVtLkRyYXdpbmcuR3JhcGhpY3NVbml0AQAAAAd2YWx1ZV9fAAgCAAAAAgAAAAs=]]></TimerFont>
```
wooferzfg, line 26: I think `TimerFont` may already be in pixels before your change?

wooferzfg:
## Before this PR
### Default Layout
Timer font: 43.75 pixel
Times font: 12 Point (equivalent to 16 pixels)
Text font: 12 Point (equivalent to 16 pixels)

### Standard layout settings factory
Timer font: 43.75 pixels
Times font: 13 pixels
Text font: 13 pixels

## After this PR
### Default Layout
**Timer font: 58.333333 pixels**
Times font: 16 pixels
Text font: 16 pixels


So the change to the Timer font in the Default Layout in this PR is incorrect, please revert that. It would also be better to update the standard layout settings factory to use 16 pixels instead of 13 pixels.

wooferzfg:
For the Timer font, the code currently scales the font down for the custom font dialog, and then scales it back after: https://github.com/LiveSplit/LiveSplit/blob/56f61c457630ff7283e9bb307961a4d49c7e1c3c/src/LiveSplit.View/View/LayoutSettingsControl.cs#L138-L150

The reason for this conversion is because the default timer font size is almost 50 pixels, and that doesn't give the user granular control since the only [size options near that size are 36, 48, 72](https://github.com/LiveSplit/LiveSplit/blob/master/lib/CustomFontDialog/CustomFontDialog/FontDialog.cs#L202-L204). If we switched to pixels, we'd need to make the size options more user friendly. However, I think this conversion isn't really causing any issues, so this PR can be merged without anything being changed there.

PR #2479, opened 2024-04-21
Move build folder cleanup to post build event

### LINKED ISSUE #2436: SpeedrunComSharp.dll is getting copied into the components folder when building locally
https://github.com/LiveSplit/LiveSplit/pull/2417#issuecomment-2038972586

wooferzfg:
Rather than cleaning up the build folder only in CI, this PR updates `LiveSplit.csproj` to handle these cleanup steps.

Resolves #2436


just-ero:
An MSBuild `Target` may be better and more maintainable.

just-ero:
Here's an example:

`_RemoveDuplicateAssemblies.targets`
```xml
<Project>

  <Target Name="AfterBuild">
    <ItemGroup>
      <_DuplicateAssemblies Include="$(OutDir)\Components\*.dll" />
    </ItemGroup>

    <Delete Files="@(_DuplicateAssemblies)"
            Condition="Exists('$(OutDir)\%(Filename)%(Extension)')" />
  </Target>

</Project>
```

`LiveSplit.csproj`
```xml
﻿<Project Sdk="Microsoft.NET.Sdk">

  <!-- ... -->

  <Import Project="_RemoveDuplicateAssemblies.targets" />

</Project>
```

LiveSplit.sln
```diff
@@ -74,6 +74,35 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{A20B2CD3-D574-467F-B1CB-CB723C3314B8}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "LiveSplit", "src\LiveSplit\LiveSplit.csproj", "{A6164B47-3C1C-472D-8DE5-CECF10619F2D}"
+	ProjectSection(ProjectDependencies) = postProject
```
just-ero, line 77: Why not make these actual `ProjectReferences` in the project file?

wooferzfg: The project doesn't actually depend on them, I just need to adjust the build order. That also sounds like a bigger change that might have its own complications, not sure

wooferzfg:
> An MSBuild Target may be better and more maintainable.

I'm not familiar enough with this, so I made a new issue: #2481

PR #2482, opened 2024-04-22
Introduce `Directory.Build.props` to collect repeating properties into one file

just-ero:
### Description

Introduces the use of `Directory.Build.props` to apply properties to all projects in the file's respective sub-directories.

- Provides common directories as properties.
- Uses .NET 8.0's artifacts output.

Need to PR the other repos as well, will open those PRs later.

### DIFF
--- a/src/LiveSplit/LiveSplit.csproj
+++ b/src/LiveSplit/LiveSplit.csproj
@@ -3,12 +3,9 @@
   <PropertyGroup>
     <ApplicationIcon>Icon.ico</ApplicationIcon>
 
-    <OutDir>..\..\bin\$(Configuration)</OutDir>
     <OutputType>WinExe</OutputType>
-    <TargetFramework>net4.6.1</TargetFramework>
     <UseWindowsForms>true</UseWindowsForms>
 
-    <Nullable>disable</Nullable>
     <PlatformTarget>AnyCPU</PlatformTarget>
   </PropertyGroup>
 
--- a/src/LiveSplit/LiveSplit.csproj
+++ b/src/LiveSplit/LiveSplit.csproj
@@ -15,16 +15,16 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\LiveSplit.Core\LiveSplit.Core.csproj" />
-    <ProjectReference Include="..\LiveSplit.View\LiveSplit.View.csproj" />
-    <ProjectReference Include="..\..\lib\WinForms Color Picker\WinFormsColor\WinFormsColor.csproj" />
+    <ProjectReference Include="$(SrcPath)\LiveSplit.Core\LiveSplit.Core.csproj" />
+    <ProjectReference Include="$(SrcPath)\LiveSplit.View\LiveSplit.View.csproj" />
+    <ProjectReference Include="$(LibPath)\WinForms Color Picker\WinFormsColor\WinFormsColor.csproj" />
   </ItemGroup>
 
   <PropertyGroup>
     <PostBuildEvent>
       mkdir "$(OutDir)\Resources"
-      cp "$(MSBuildProjectDirectory)\..\..\res\SplitsFile.ico" "$(OutDir)\Resources"
-      cp "$(MSBuildProjectDirectory)\..\..\res\LayoutFile.ico" "$(OutDir)\Resources"
+      cp "$(ResourcesPath)\SplitsFile.ico" "$(OutDir)\Resources"
+      cp "$(ResourcesPath)\LayoutFile.ico" "$(OutDir)\Resources"
       rd /s /q "$(OutDir)\runtimes"
       del "$(OutDir)\Microsoft.WindowsAPICodePack.ExtendedLinguisticServices.dll"
       del "$(OutDir)\Microsoft.WindowsAPICodePack.Sensors.dll"
--- a/src/LiveSplit/LiveSplit.csproj
+++ b/src/LiveSplit/LiveSplit.csproj
@@ -23,8 +23,8 @@
   <PropertyGroup>
     <PostBuildEvent>
       mkdir "$(OutDir)\Resources"
-      cp "$(ResourcesPath)\SplitsFile.ico" "$(OutDir)\Resources"
-      cp "$(ResourcesPath)\LayoutFile.ico" "$(OutDir)\Resources"
+      copy "$(ResourcesPath)\SplitsFile.ico" "$(OutDir)\Resources"
+      copy "$(ResourcesPath)\LayoutFile.ico" "$(OutDir)\Resources"
       rd /s /q "$(OutDir)\runtimes"
       del "$(OutDir)\Microsoft.WindowsAPICodePack.ExtendedLinguisticServices.dll"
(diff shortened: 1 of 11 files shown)


src/LiveSplit/LiveSplit.csproj
```diff
@@ -18,20 +15,23 @@
...
 
   <PropertyGroup>
     <PostBuildEvent>
       mkdir "$(OutDir)\Resources"
-      cp "$(MSBuildProjectDirectory)\..\..\res\SplitsFile.ico" "$(OutDir)\Resources"
-      cp "$(MSBuildProjectDirectory)\..\..\res\LayoutFile.ico" "$(OutDir)\Resources"
+      copy "$(ResourcesPath)\SplitsFile.ico" "$(OutDir)\Resources"
+      copy "$(ResourcesPath)\LayoutFile.ico" "$(OutDir)\Resources"
       rd /s /q "$(OutDir)\runtimes"
+      rd /s /q "$(OutDir)\Components\runtimes"
```
wooferzfg, line 29: Why are we removing the `Components\runtimes` directory? Isn't that required for Racetime?

just-ero: In my own local tests, a `$(OutDir)\Components\WebView2Loader.dll` file was generated. This appears to be the same file from the `runtimes/` folder. LiveSplit and Racetime appear to function normally without the directory.

wooferzfg: Are you able to click on a race and view the race entrants & chat without that file present?

just-ero: I will have to test it later. I do not have a Racetime account.

wooferzfg: > In my own local tests, a $(OutDir)\Components\WebView2Loader.dll file was generated.

I don't see that file in the CI artifact

just-ero: Ah. Locally, I pass `--use-current-runtime` to `dotnet build`. I did it because I wasn't fond of the assemblies in `runtimes/win-x86` and `runtimes/win-arm64`. Looks like specifying the target runtime also has this build behavior as a side effect.

just-ero: How would you like to continue with this? Should I add `--ucr` to the CI, or should I not remove `Components/runtimes`?

wooferzfg: Also, if you use `--use-current-runtime`, does that still put files in `Components\runtimes`?

wooferzfg: Instead of setting it in CI, we could maybe set the `RuntimeIdentifier` property somewhere https://learn.microsoft.com/en-us/dotnet/core/rid-catalog#windows-rids

But we're already shipping 32 bit and 64 bit versions of livesplit-core - is this any different?

just-ero: I don't think it should matter. We can just leave it like it was before and I remove that line.

I would prefer to specify as few properties as possible in the projects themselves. It allows the user to customize their build as much as they want. Especially setting `OutDir` is a thorn in my eye; I'd prefer setting `-o` on the command line. I'm still working on a solution there.

wooferzfg: > I would prefer to specify as few properties as possible in the projects themselves. It allows the user to customize their build as much as they want.

This is pretty much the polar opposite of what I have in mind for the build process. I'd like my local build to match the CI build, so that I can do development and debugging knowing that it will be the same when it's eventually released. I'm not sure why you'd need to customize the build when the application will always be shipped in a particular way.

wooferzfg: The output directory is fine to keep on the command line, though, since that doesn't affect the behavior of the application.

just-ero: I would want the user to be able to compile for whatever runtime they want, for whatever reason they want. If they just want to test something on win-arm64, who are we to force a runtime?

You said it best yourself, there's a specific result you expect when shipping the app. But locally, you *aren't* shipping. It's a privilege to even get the build output users are getting now.

This would be more appropriate for something like `dotnet publish`. This is where I would expect the build process to go the extra mile and actually add all the components and stuff. But I also understand that when you simply run the `LiveSplit.csproj` project, one would expect it to be the full app (which is obviously not the case, the `LiveSplit.csproj` project doesn't inherently have anything to do with components or registering).

wooferzfg: > If they just want to test something on win-arm64, who are we to force a runtime?

Isn't it possible to override the properties set in the `.csproj` files by using command line arguments?

wooferzfg: > It's a privilege to even get the build output users are getting now.

I think the bare minimum is that everything works correctly when building & running locally

just-ero: What *is* "correctly", though? `LiveSplit.exe` should technically run without any components being present. This is currently not the case, but that can be changed. You need at least `LiveSplit.Timer` present.

Let's look at it from another perspective even: I do not use Visual Studio. I use VSCode and the .NET CLI. If I run `dotnet run ./src/LiveSplit`, I get an error, since none of the components were built.

wooferzfg: > What is "correctly", though?

The behavior of the application should be the same as the release build. It's fine if there's small differences like `.pdb` files or it running slower since it's a debug build, but in general I should be able to run the app and test it with the expectation that it will work the same when it's released.

> LiveSplit.exe should technically run without any components being present. This is currently not the case, but that can be changed. You need at least LiveSplit.Timer present.

LiveSplit always ships with the standard components, and "Timer Only Mode" is part of the app, so the components should be built before the program is run.

wooferzfg: >  We can just leave it like it was before and I remove that line.

Can we go with this approach?

just-ero: > I want the dev environment to match production, at least with the "default" way to build & run the program

`dotnet run` *is* my default way to run the program, and it does *not* match "production".

wooferzfg: That sounds like a potential improvement, then

just-ero: I still strongly disagree with forcibly making the build output the exact same thing the CI might.

Think about it. All of the projects exist isolated from one another. Some have references to another, but that's for compilation reasons, of course. There's no reason and no indication that running the `LiveSplit.csproj` project would build and move the components as well, which are all their own projects as well, existing isolated from the app itself.  
"Building" means something very specific in .NET.

The behavior you want is reserved for something like publishing the app. The name already implies that something more than just building is happening.

---

Like I already expressed, I agree that this would be a hit towards ease of use. Maybe there are ways to publish & run, but I would need to look into that.

just-ero: Especially moving the resource files (icons) is something I completely do not understand. They are not required for functionality and they're therefore only steps that slow down how much time running the app takes.

PR #2484, opened 2024-04-25
Use MSBuild targets for build

### LINKED ISSUE #2481: Use an MSBuild Target instead of PostBuildEvent for cleaning up build folder
Suggestion: https://github.com/LiveSplit/LiveSplit/pull/2479#issuecomment-2068179162
Example: https://github.com/LiveSplit/LiveSplit/pull/2479#issuecomment-2068180331

PostBuildEvent to replace:
https://github.com/LiveSplit/LiveSplit/blob/6f9cf77961c714522e054488a0d2c59cfa2a3f63/src/LiveSplit/LiveSplit.csproj#L26-L43

Target element docs: https://learn.microsoft.com/en-us/visualstudio/msbuild/target-element-msbuild?view=vs-2022

ero-qt:
### Description

- Drops all of the hard-coded `OutDir` properties.
- Moves pre and post-build events to MSBuild targets.

### DIFF
--- a/src/LiveSplit.Core/LiveSplit.Core.csproj
+++ b/src/LiveSplit.Core/LiveSplit.Core.csproj
@@ -37,21 +37,6 @@
     <Reference Include="WindowsBase" />
   </ItemGroup>
 
-  <PropertyGroup>
-    <PreBuildEvent>
-      mkdir "$(MSBuildProjectDirectory)\Updates\GitInfo\."
-      where git
-      if %25ERRORLEVEL%25 == 0 (
-        git describe --dirty --always --long --tags &gt; "$(MSBuildProjectDirectory)\Updates\GitInfo\version.txt"
-        git rev-parse --abbrev-ref HEAD &gt; "$(MSBuildProjectDirectory)\Updates\GitInfo\branch.txt"
-        git rev-parse HEAD &gt; "$(MSBuildProjectDirectory)\Updates\GitInfo\revision.txt"
-      ) else (
-        touch "$(MSBuildProjectDirectory)\Updates\GitInfo\version.txt"
-        touch "$(MSBuildProjectDirectory)\Updates\GitInfo\branch.txt"
-        touch "$(MSBuildProjectDirectory)\Updates\GitInfo\revision.txt"
-      )
-      EXIT 0
-    </PreBuildEvent>
-  </PropertyGroup>
+  <Import Project="$(MSBuildThisFileDirectory)\targets\_CreateGitInfo.targets" />
 
 </Project>
--- a/src/LiveSplit.Core/LiveSplit.Core.csproj
+++ b/src/LiveSplit.Core/LiveSplit.Core.csproj
@@ -37,6 +37,6 @@
     <Reference Include="WindowsBase" />
   </ItemGroup>
 
-  <Import Project="$(MSBuildThisFileDirectory)\targets\_CreateGitInfo.targets" />
+  <Import Project="$(MSBuildThisFileDirectory)\targets\*.targets" />
 
 </Project>
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Set up .NET environment
         uses: actions/setup-dotnet@v4
         with:
-          dotnet-version: '8.x'
+          dotnet-version: '8.0.x'
 
       - name: Build LiveSplit
         run: >
@@ -36,7 +36,8 @@ jobs:
           -v m
           -c Release
           -p:DebugType=None
-          .\LiveSplit.sln
+          -o .\ci-build
+          .\src\LiveSplit
 
       - name: Build UpdateManager.exe
         run: >
@@ -44,7 +45,8 @@ jobs:
           -v m
           -c UpdateManagerExe
           -p:DebugType=None
-          .\src\UpdateManager\UpdateManager.csproj
+          -o .\ci-update-manager
+          .\src\UpdateManager
 
       - name: Run tests
         run: >
@@ -58,18 +60,18 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: LiveSplit_Build
-          path: .\bin\Release
+          path: .\ci-build
 
       - name: Upload UpdateManager.exe as an artifact
         uses: actions/upload-artifact@v4
         with:
           name: UpdateManagerExe
-          path: .\bin\UpdateManagerExe
+          path: .\ci-update-manager
 
       - name: Upload build to LiveSplit.github.io
         if: github.repository == 'LiveSplit/LiveSplit' && github.ref == 'refs/heads/master'
         run: |
-          7z a LiveSplitDevBuild.zip .\bin\Release\*
+          7z a LiveSplitDevBuild.zip .\ci-build\*
           git config --global user.email "action@github.com"
           git config --global user.name "GitHub Action"
           git clone -q --branch master --single-branch "https://github.com/LiveSplit/LiveSplit.github.io.git"
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -20,8 +20,6 @@
   <PropertyGroup Label="Output Settings">
     <UseArtifactsOutput>true</UseArtifactsOutput>
     <ArtifactsPath>$(RootPath)\artifacts</ArtifactsPath>
-
-    <BuildPath>$(RootPath)\bin\$(Configuration.ToLowerInvariant())</BuildPath>
   </PropertyGroup>
 
 </Project>
--- a/components/Directory.Build.props
+++ b/components/Directory.Build.props
@@ -7,8 +7,4 @@
     <LsLibPath>$(RootPath)\lib</LsLibPath>
   </PropertyGroup>
 
-  <PropertyGroup Label="Output Settings">
-    <OutDir>$(BuildPath)\Components</OutDir>
-  </PropertyGroup>
-
 </Project>
--- a/lib/CustomFontDialog/CustomFontDialog/CustomFontDialog.csproj
+++ b/lib/CustomFontDialog/CustomFontDialog/CustomFontDialog.csproj
@@ -1,7 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <OutDir>bin\$(Configuration)</OutDir>
     <TargetFramework>net4.6.1</TargetFramework>
     <UseWindowsForms>true</UseWindowsForms>
 
--- a/lib/WinForms Color Picker/WinFormsColor/WinFormsColor.csproj
+++ b/lib/WinForms Color Picker/WinFormsColor/WinFormsColor.csproj
@@ -1,7 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <OutDir>bin\$(Configuration)</OutDir>
     <TargetFramework>net4.6.1</TargetFramework>
     <UseWindowsForms>true</UseWindowsForms>
 
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,9 +1,0 @@
-<Project>
-
-  <Import Project="..\Directory.Build.props" />
-
-  <PropertyGroup Label="Output Settings">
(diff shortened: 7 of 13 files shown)


just-ero:
> Could not find a part of the path 'D:\a\LiveSplit\LiveSplit\src\LiveSplit.Core\Updates\GitInfo\branch.txt

Works fine locally.

src/LiveSplit.Core/LiveSplit.Core.csproj
```diff
@@ -37,21 +37,6 @@
...
-        git rev-parse HEAD &gt; "$(MSBuildProjectDirectory)\Updates\GitInfo\revision.txt"
-      ) else (
-        touch "$(MSBuildProjectDirectory)\Updates\GitInfo\version.txt"
-        touch "$(MSBuildProjectDirectory)\Updates\GitInfo\branch.txt"
-        touch "$(MSBuildProjectDirectory)\Updates\GitInfo\revision.txt"
-      )
-      EXIT 0
-    </PreBuildEvent>
-  </PropertyGroup>
+  <Import Project="_CreateGitInfo.targets" />
```
just-ero, line 40: It's as if this project is not actually being imported?

just-ero:
Everything appears to be working now.

It's worth noting that in the CI, the `LiveSplit.Core` pre-build event runs twice now. I'm not sure why that might be.

just-ero:
Indeed. I don't know what causes this, the fact they are `ProjectReference`s should prevent `LiveSplit.Core` from being built more than once.

If you feel like losing your mind, you can analyze the build process using an MSBuild log:
```pwsh
> dotnet build -bl 'src/LiveSplit'
```

You can view the generated `msbuild.binlog` file using the MSBuild structured log viewer: https://msbuildlog.com.

wooferzfg:
When I ran "Rebuild Solution" in Visual Studio, it didn't copy the component DLLs to the `Components` folder. `dotnet build LiveSplit.sln` did work, however.

just-ero:
Ah, we should remove the project dependencies from the solution file. This won't affect anything pertaining to the build however.

just-ero:
I believe I know what may be causing the secondary `LiveSplit.Core` build: we set `Properties` in the `ProjectReference` for the components:
> MSBuild loads a unique Node per unique set of global properties. So, altering the global properties means you are kicking off a second node of the same thing. Which will run at the same time potentially as another copy of the same. And thus clobber or step on files, or lock things.

(someone from the .NET Evolution Discord server)

wooferzfg:
> Ah, we should remove the project dependencies from the solution file. This won't affect anything pertaining to the build however.

Does the CI build work without the project dependencies? When we previously didn't have them, `LiveSplit.csproj` wasn't guaranteed to build last, which made the post build event not always have the files it needs at the time it was ran.

Also, I'd definitely like rebuilding the solution to move the component files to `Components`, since using Visual Studio will be much more inconvenient otherwise.

just-ero:
> I'd definitely like rebuilding the solution to move the component files to `Components`

Building `LiveSplit.csproj` (which should happen when the entire solution is built, obviously) should also build all of the components and therefore also put them into `$(OutputPath)/Components`.

Let me propose another solution: build the components without passing another property and instead *copy* the files resulting from the build. Unfortunately, I have absolutely no idea how to capture all of the outputs, only the immediate result (for example `LiveSplit.Racetime.dll` instead of `LiveSplit.Racetime.dll;DarkUI.dll;...`.

Here's how that might look:
```xml
<ItemGroup>
  <ProjectReference Include="$(ComponentsPath)\*\src\*\*.csproj"
                    OutputItemType="ComponentsOutput"
                    ReferenceOutputAssembly="false" />
</ItemGroup>

<Target Name="_CopyComponentsOutput" AfterTargets="Build">
  <Copy Files="@(ComponentsOutput)" DestinationFolder="$(OutputPath)\Components" />
</Target>
```

just-ero:
I'm honestly still not really following. Why does building the solution matter now? It's not exactly meant to create any sort of "production output". I will reiterate once again that the *only* step that should produce the output you desire is a publish.

Building `LiveSplit.csproj` produces the output you want now, which is already way more than a build should handle.

> The point of using the Publish output is it's really the only thing that integrates with the full lifecycle of the .NET SDK assembly generation process. For instance, if you enable R2R on the plugin assemblies, that only happens during Publish.

> "Build" output isn't really guaranteed to be a distributable unit, but merely something capable of being run in a debugger, etc. At least that's how the new SDK treats it.

just-ero:
Furthermore, I cannot replicate your problem. `dotnet build LiveSplit.sln` builds the solution and places the output correctly in `./artifacts/bin/LiveSplit/<configuration>`.

wooferzfg:
> I'm honestly still not really following. Why does building the solution matter now? It's not exactly meant to create any sort of "production output". I will reiterate once again that the only step that should produce the output you desire is a publish.

I need an output that allows LiveSplit to run properly without me needing to manually move files around. If there's no `Components` folder, then LiveSplit crashes on startup.

> Building LiveSplit.csproj produces the output you want now, which is already way more than a build should handle.
> Furthermore, I cannot replicate your problem. 

As I [mentioned above](https://github.com/LiveSplit/LiveSplit/pull/2484#issuecomment-2077182398), it only happens in Visual Studio. Regardless of whether I rebuild the solution or just build LiveSplit.csproj, the `Components` directory does not get created within `artifacts\bin\LiveSplit\debug`.

just-ero:
Hey @wooferzfg, I've done some more work here. Let me know if there's anything still incorrect.

We should also remove the `OutDir` property in the `SpeedrunComSharp` project, I'll see if I feel like opening a PR there.

wooferzfg:
Looks like the build is currently failing

wooferzfg:
I just re-ran the build on `master` and it passed https://github.com/LiveSplit/LiveSplit/actions/runs/9390637185/job/27054550614

just-ero:
I believe it may be the use of the `MSBuild` task, which is really weird. Must be an issue on Microsoft's side, maybe it's intended behavior even...

just-ero:
At this point I don't even know how to fix this? There's not an official CLI installer for this version of the dev pack.

wooferzfg:
Would it work if we upgraded to 4.6.2?

just-ero:
Let me know if there's anything else you want me to change. :)

wooferzfg:
It looks like `LiveSplit.Register.exe` and `LiveSplit.Register.exe.config` are not being included in the build. The rest looks good to me

just-ero:
That should do it!

wooferzfg:
The CI build looks good to me now. However, when doing "Rebuild Solution" in Visual Studio, the `Components` folder is still empty, and `LiveSplit.Register.exe` is missing. I get a bunch of errors at the end that look like this:
```
35>Done building project "LiveSplit.Timer.csproj" -- FAILED.
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\CustomFontDialog.dll' could not be found
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\LiveSplit.Core.dll' could not be found
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\SpeedrunComSharp.dll' could not be found
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\UpdateManager.dll' could not be found
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\WinFormsColor.dll' could not be found
35>Done building project "LiveSplit.Title.csproj" -- FAILED.
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\CustomFontDialog.dll' could not be found
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\LiveSplit.Core.dll' could not be found
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\SpeedrunComSharp.dll' could not be found
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\UpdateManager.dll' could not be found
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\WinFormsColor.dll' could not be found
35>Done building project "LiveSplit.TotalPlaytime.csproj" -- FAILED.
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\CustomFontDialog.dll' could not be found
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\LiveSplit.Core.dll' could not be found
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\SpeedrunComSharp.dll' could not be found
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\UpdateManager.dll' could not be found
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\WinFormsColor.dll' could not be found
35>Done building project "LiveSplit.Video.csproj" -- FAILED.
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\CustomFontDialog.dll' could not be found
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\LiveSplit.Core.dll' could not be found
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\SpeedrunComSharp.dll' could not be found
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\UpdateManager.dll' could not be found
35>CSC : error CS0006: Metadata file 'C:\Users\user\Documents\GitHub\LiveSplit\artifacts\bin\LiveSplit\debug\Components\WinFormsColor.dll' could not be found
35>Done building project "LiveSplit.WorldRecord.csproj" -- FAILED.
========== Rebuild All: 34 succeeded, 1 failed, 0 skipped ==========
========== Rebuild completed at 8:45 AM and took 17.341 seconds ==========
```

just-ero:
Reconvening with the MSBuild team. To me it seems like a Visual Studio bug; it doesn't even generate the `artifacts` directory. Maybe an issue with using `Directory.Build.props`?

just-ero:
> Are these projects in your solution file?
> That's the problem. Whatever is the default build configuration in the solution file is including those projects, and so VS is attempting to build them.
> You're working with a pretty exotic project layout here, so I'm not sure you're going to be able to force VS to behave the way you want it to.
> The thing I would suggest, is to try and make two different solution configurations in the VS UI - in the first include only your core. In the second include your sub module components. In this way you can at least control the ordering to a greater degree.

This sounds unsolvable to me.

wooferzfg:
There's two paths forward then:
- Closing this PR (and reverting the SpeedrunComSharp change?)
- No longer supporting building the solution in Visual Studio, and updating the README to indicate this

I don't have a good understanding of MSBuild targets - what are we gaining with this approach?

wooferzfg:
Should I revert the SpeedrunComSharp change for now, or does that still work independently of this PR?

just-ero:
You will have to revert the change. I apologize.

wooferzfg:
My fault for merging before testing with VS again

just-ero:
Oh, great.

PR #2485, opened 2024-04-25
Add share to X (Twitter) with web intents.

Kuhaku9893:
Share to X (Twitter) using web intents instead of API.
When sharing, a screenshot is copied to the clipboard, so you can use ctrl+v on the post edit screen to paste the screenshot.

### DIFF
--- a/src/LiveSplit.Core/LiveSplit.Core.csproj
+++ b/src/LiveSplit.Core/LiveSplit.Core.csproj
@@ -37,6 +37,21 @@
     <Reference Include="WindowsBase" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Update="Web\Share\ShareSettings.Designer.cs">
+      <DesignTimeSharedInput>True</DesignTimeSharedInput>
+      <AutoGen>True</AutoGen>
+      <DependentUpon>ShareSettings.settings</DependentUpon>
+    </Compile>
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Update="Web\Share\ShareSettings.settings">
+      <Generator>SettingsSingleFileGenerator</Generator>
+      <LastGenOutput>ShareSettings.Designer.cs</LastGenOutput>
+    </None>
+  </ItemGroup>
+
   <PropertyGroup>
     <PreBuildEvent>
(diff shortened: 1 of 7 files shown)


just-ero:
In the future, for substantial changes like this which affect LiveSplit's feature set, please open an issue first.

src/LiveSplit.Core/LiveSplit.Core.csproj
```diff
@@ -37,6 +37,21 @@
...
+      <DependentUpon>ShareSettings.settings</DependentUpon>
+    </Compile>
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Update="Web\Share\ShareSettings.settings">
+      <Generator>SettingsSingleFileGenerator</Generator>
+      <LastGenOutput>ShareSettings.Designer.cs</LastGenOutput>
+    </None>
+  </ItemGroup>
```
wooferzfg, line 53: Do we need this? @just-ero do you know what this does?

just-ero: I can't tell you, sorry. This appears to have been generated by Visual Studio. VSCode does not have a designer, so generating a settings file like this is not even possible there.

We can leave it in for now and I'll do more research.

src/LiveSplit.Core/Web/Share/Twitter.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        public static readonly Uri BaseUri = new Uri("https://twitter.com/intent/tweet");
+
+        protected Twitter() { }
+
+        public string PlatformName => "X (Twitter)";
+
+        public string Description =>
+@"X (Twitter) allows you to share your run with the world. 
+When sharing, a screenshot of LiveSplit is automatically copied the the clipboard.
```
just-ero, line 27:
```suggestion
When sharing, a screenshot of LiveSplit is automatically copied to the clipboard.
```

src/LiveSplit.Core/LiveSplit.Core.csproj
wooferzfg: I'd rather remove it since the build works without it

PR #2504, opened 2024-06-04
Lua Client Documentation - First Commit

CRGorman:
Functional Script for the [FSOpen](https://github.com/scp-fs2open/fs2open.github.com) project available [here](https://github.com/FSO-Scripters/fso-scripts/pull/66/files).

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,8 @@ public class MainTest {
 ```
 
 #### Lua
-Lua is usable for a client, however the lua io library must be available for the script to use.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -177,6 +177,17 @@ public class MainTest {
 }
 ```
 
+#### Lua
+Lua is usable for a client, however the lua io library must be available for the script to use.
```
wooferzfg, line 181: Can we link to the lua io library here?

Also, a grammar fix:
```suggestion
Lua is usable for a client. However, the lua io library must be available for the script to use.
```

CRGorman: Lua io is part of the standard package. Depending on the implementation though, you can manually load the set and skip over IO entirely. [See here for the specifics](https://www.lua.org/manual/5.3/manual.html#6).

PR #2513, opened 2024-07-17
WebSocket Server

TheTedder:
This PR incorporates the WebSocket server into LiveSplit, refactoring some of the connection code to allow for connection types that don't inherit from `Connection`. The only functional difference between this server and the TCP server is that it uses discrete messages instead of splitting commands with newline characters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,17 +68,17 @@ The documentation for how to develop, test, and submit an Auto Splitter can be f
 
 ## The LiveSplit Server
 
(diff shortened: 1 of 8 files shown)


CryZe:
The only thing I am concerned about is that this might cause confusion with the websocket server support in LSO, which is not only not compatible with this, but also is a different JSON based protocol.

wooferzfg:
I think we've had this problem for a while, though - the server in LS is still mostly the same as the original Server component implementation, while the server in LSO is different

wooferzfg:
I'll merge this for now since this is consistent with the TCP server, but let's try to figure out how to make things less confusing/inconsistent between LS and LSO before the next release

PR #2518, opened 2024-08-16
Upgrade to .NET Framework 4.8.1 (+ some clean-up)

just-ero:
### Description

This PR upgrades LiveSplit's target framework to .NET Framework 4.8.1, the most recent stable release.

The PR also includes some clean-up, and I apologize for that. I thought it may be less annoying having to merge just one PR here and on all of the components instead of opening multiple ones. I will revert those additional commits if so desired.


just-ero:
There may be certain components which cannot be activated due to this change. One example I found is SoulSplitter. A lot of other components appear to still be compatible.

wooferzfg:
I think that's a problem - we don't want to break any existing components (the existing released component DLLs will need to continue working after any change we make)

just-ero:
That would mean we can never upgrade the target framework away from the current one which has been deprecated since 2022. I don't see this as an option.

wooferzfg:
> That would mean we can never upgrade the target framework away from the current one which has been deprecated since 2022. I don't see this as an option.

4.6.2 hasn't reached EOL yet (scheduled for 2027), so that's another potential route

wooferzfg:
Could you update the README? https://github.com/LiveSplit/LiveSplit?tab=readme-ov-file#compiling

README.md
```diff
@@ -47,9 +47,9 @@ You can browse the [Issues](https://github.com/LiveSplit/LiveSplit/issues) to fi
 
 ## Compiling
 
-LiveSplit uses .NET Framework 4.6.1. To compile LiveSplit, you need the following components installed:
+LiveSplit uses .NET Framework 4.8.1. To compile LiveSplit, you need the following components installed:
 - [.NET 8.0 SDK](https://dotnet.microsoft.com/en-us/download/dotnet/8.0)
-- [.NET Framework 4.6.1 Developer Pack](https://dotnet.microsoft.com/en-us/download/dotnet-framework/net461)
+- [.NET Framework 4.8.1 Developer Pack](https://dotnet.microsoft.com/en-us/download/dotnet-framework/thank-you/net481-developer-pack-offline-installer)
```
just-ero, line 52: Why not just `https://dotnet.microsoft.com/en-us/download/dotnet-framework/net481` like before?

wooferzfg: https://dotnet.microsoft.com/en-us/download/dotnet-framework/net481 has a link to the runtime, but the link title says "developer pack" specifically

just-ero: image

wooferzfg: Yes, I changed it to the developer pack link specifically

just-ero: Didn't see the edit. :P

wooferzfg:
I did a fresh clone + rebuild solution in Visual Studio 2022, but I'm getting these errors:

devenv_tP8XSRnxcV

```
System.IO.DirectoryNotFoundException: Could not find a part of the path 'C:\Users\spapu\Documents\GitHub\LiveSplit\props\artifacts\obj\LiveSplit\debug\LiveSplit.exe.withSupportedRuntime.config'.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize)
   at System.Xml.XmlDownloadManager.GetStream(Uri uri, ICredentials credentials, IWebProxy proxy, RequestCachePolicy cachePolicy)
   at System.Xml.XmlUrlResolver.GetEntity(Uri absoluteUri, String role, Type ofObjectToReturn)
   at System.Xml.XmlTextReaderImpl.FinishInitUriString()
   at System.Xml.XmlReaderSettings.CreateReader(String inputUri, XmlParserContext inputContext)
   at System.Xml.Linq.XDocument.Load(String uri, LoadOptions options)
   at Microsoft.Build.Tasks.GenerateBindingRedirects.LoadAppConfig(ITaskItem appConfigItem)
   at Microsoft.Build.Tasks.GenerateBindingRedirects.Execute()
   at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute()
   at Microsoft.Build.BackEnd.TaskBuilder.<ExecuteInstantiatedTask>d__26.MoveNext()	LiveSplit	C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Current\Bin\amd64\Microsoft.Common.CurrentVersion.targets	2509	
```

just-ero:
Did you install the .NET Framework 4.8.1 developer pack?

wooferzfg:
Yes I did

wooferzfg:
I don't get any errors if I revert the changes in the 3 cleanup commits

just-ero:
There appears to be an issue on your end. It seems to place an `artifacts` folder within the `props` folder, which is not expected behavior.

I haven't experienced this personally and wouldn't know why you're experiencing it.

wooferzfg:
There's no guarantee that this issue won't come up for someone else, though, so I'd like to solve it prior to merging

just-ero:
No, you don't understand. This cannot happen. There is nothing that says an `artifacts` directory should be created inside of `props`.

wooferzfg:
But it *is* happening

src/LiveSplit/app.manifest
```diff
@@ -1,25 +1,6 @@
...
-      <!-- Windows 8 -->
-      <supportedOS Id="{4a2f28e3-53b9-4441-ba9c-d69d4a4a6e38}" />
-
-      <!-- Windows 8.1 -->
-      <supportedOS Id="{1f676c76-80e1-4239-95bb-83d0f6d0da78}" />
-
-      <!-- Windows 10+ -->
-      <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}" />
-    </application>
-  </compatibility>
```
wooferzfg, line 21: Why are these being removed?

just-ero: They are not required for the DPI-awareness feature to work. I wanted to push this change in the previous PR, but you were too quick to merge it. 😅

wooferzfg:
Changing `RootPath` to `<RootPath>$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), LiveSplit.sln))</RootPath>` fixes the error I was describing. I'm guessing that the `\..` being in the path breaks in some places

just-ero:
Try using `$([System.IO.Path]::GetDirectoryName($(MSBuildThisFileDirectory))` instead.

But still, I don't believe that should cause a problem...

Could you also try simply removing the `\` in the original property? So `$(MSBuildThisFileDirectory)..`?

Could you also see which .NET SDK you have installed? `dotnet --list-sdks`.

wooferzfg:
> Try using `$([System.IO.Path]::GetDirectoryName($(MSBuildThisFileDirectory))` instead.

I did `<RootPath>$([System.IO.Path]::GetDirectoryName($(MSBuildThisFileDirectory)))</RootPath>` and it put everything into `props`.

> Could you also try simply removing the `\` in the original property? So `$(MSBuildThisFileDirectory)..`?

That does build correctly

> Could you also see which .NET SDK you have installed? `dotnet --list-sdks`.

```
6.0.406 [C:\Program Files\dotnet\sdk]
8.0.202 [C:\Program Files\dotnet\sdk]
```

just-ero:
What operating system are you on?

wooferzfg:
Windows 11

wooferzfg:
Should we change it to `<RootPath>$(MSBuildThisFileDirectory)..</RootPath>`?

just-ero:
We can do that for now, but we may want to check if we do this (or something similar) anywhere else that might need adjusting.

wooferzfg:
I'm not sure what causes the difference, but `\..` seems incorrect since [the docs](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-reserved-and-well-known-properties?view=vs-2022) mention that `MSBuildThisFileDirectory` includes the final backslash.

just-ero:
I suppose. I just like adding the extra backslash for consistency/readability.

just-ero:
@wooferzfg Can you take a look at this conflict? I can't figure it out.

wooferzfg:
Your branch in `LiveSplit.Timer` is 1 commit behind `master`

wooferzfg:
I think you could also merge `master` into this PR and then update the commit in `LiveSplit.Timer` so that it points to yours

PR #2520, opened 2024-08-16
Update ShortTimeFormatter.cs

10-days-till-xmas:
Allows for 3 decimal places for use in components


wooferzfg:
This change affects more than just components - it updates the default value to milliseconds

wooferzfg:
Why can't a component just use `GeneralTimeFormatter`?

10-days-till-xmas:
ah i didnt know abt  `GeneralTimeFormatter`, I'll use that instead

10-days-till-xmas:
Would it be okay to make a new `MillisecondTimeFormatter` file instead

wooferzfg:
You shouldn't need that for the implementation. All the components with an accuracy setting already pass it to the time formatter, instead of using a hardcoded accuracy value.

Issue #930, opened 2017-11-16
Show milliseconds in Splits Editor and in the Layout


wooferzfg:

leon42-sys:
Sadge

just-ero:
What's the status on this?

PR #2521, opened 2024-08-18
Allow the use of 3 decimal places in RunEditorDialog

10-days-till-xmas:
Relates to #930

### DIFF
--- a/src/LiveSplit.Core/TimeFormatters/ShortTimeFormatterMilliseconds.cs
+++ b/src/LiveSplit.Core/TimeFormatters/ShortTimeFormatterMilliseconds.cs
@@ -21,5 +21,6 @@ namespace LiveSplit.TimeFormatters
         public string Format(TimeSpan? time)
         {
             return Format(time, DigitsFormat.SingleDigitSeconds);
+        }
     }
 }
--- a/src/LiveSplit.Core/TimeFormatters/ShortTimeFormatterMilliseconds.cs
+++ b/src/LiveSplit.Core/TimeFormatters/ShortTimeFormatterMilliseconds.cs
@@ -10,7 +10,9 @@ namespace LiveSplit.TimeFormatters
             Accuracy = TimeAccuracy.Milliseconds;
             NullFormat = NullFormat.ZeroWithAccuracy;
         }
+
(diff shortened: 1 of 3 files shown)


src/LiveSplit.Core/TimeFormatters/ShortTimeFormatterMilliseconds.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace LiveSplit.TimeFormatters
+{
+    public class ShortTimeFormatterMilliseconds : GeneralTimeFormatter
+    {
+        public ShortTimeFormatterMilliseconds() 
+        {
+            Accuracy = TimeAccuracy.Milliseconds;
+            NullFormat = NullFormat.ZeroWithAccuracy;
+        }
+        public string Format(TimeSpan? time, DigitsFormat format)
```
wooferzfg, line 13:
```suggestion
        }

        public string Format(TimeSpan? time, DigitsFormat format)

```

src/LiveSplit.Core/TimeFormatters/ShortTimeFormatterMilliseconds.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Globalization;
+
+namespace LiveSplit.TimeFormatters
+{
+    public class ShortTimeFormatterMilliseconds : GeneralTimeFormatter
+    {
+        public ShortTimeFormatterMilliseconds() 
+        {
+            Accuracy = TimeAccuracy.Milliseconds;
+            NullFormat = NullFormat.ZeroWithAccuracy;
```
wooferzfg, line 11: Why are we setting `Accuracy` twice? (I don't think it makes sense in `ShortTimeFormatter` either)

wooferzfg: I think it would make more sense for the constructor to take the `DigitsFormat`, and then you can just get rid of the `Format` override

wooferzfg: @10-days-till-xmas This feedback is not addressed yet

PR #2528, opened 2024-08-28
Add CustomVariables

AlexKnauth:
The `LiveSplit.Core` part of changes to add Custom Variables to LiveSplit.

I also have a branch lined up with changes to `LiveSplit.AutoSplittingRuntime` to allow autosplitters to set these Custom Variables: https://github.com/LiveSplit/LiveSplit.AutoSplittingRuntime/pull/16

Tasks:
- [x] Test parsing `.lss` runs with permanent custom variables created in LiveSplit One, and saving it without modifying with an autosplitter.
  - It's fine that the LSO tag `SpeedrunComVariables` was translated to the LS tag `Variables`.
  - The important thing is that the LSO tag `CustomVariables` was both parsed and saved correctly, staying as `CustomVariables`.
- [x] Test saving `.lss` runs with permanent custom variables modified by an autosplitter.
- [x] Test parsing those modified `.lss` runs back in LiveSplit.
- [x] Test parsing those modified `.lss` runs back in LiveSplit One.
  - It's fine that the LS tag `Variables` was translated to the LSO tag `SpeedrunComVariables`.
  - The important thing is LS `CustomVariables` to LSO `CustomVariables` went smoothly.

### DIFF
--- a/src/LiveSplit.Core/Model/RunFactories/StandardFormatsRunFactory.cs
+++ b/src/LiveSplit.Core/Model/RunFactories/StandardFormatsRunFactory.cs
@@ -134,6 +134,15 @@ public class StandardFormatsRunFactory : IRunFactory
             }
         }
 
+        using (LiveSplitCore.RunMetadataCustomVariablesIter iter = metadata.CustomVariables())
+        {
+            LiveSplitCore.RunMetadataCustomVariableRef variable;
+            while ((variable = iter.Next()) != null)
+            {
+                run.Metadata.GetOrAddCustomVariable(variable.Name()).AsPermanent().Value = variable.Value();
+            }
+        }
+
         run.GameIcon = ParseImage(lscRun.GameIconPtr(), lscRun.GameIconLen());
         run.GameName = lscRun.GameName();
         run.CategoryName = lscRun.CategoryName();
--- a/src/LiveSplit.Core/Model/RunMetadata.cs
+++ b/src/LiveSplit.Core/Model/RunMetadata.cs
@@ -119,6 +119,64 @@ public class RunMetadata
         }
     }
 
+    /// <summary>
+    ///     A dictionary mapping custom variable names to <see cref="CustomVariable"/> objects.
+    /// </summary>
+    public Dictionary<string, CustomVariable> CustomVariables { get; private set; } = new Dictionary<string, CustomVariable>();
+
+    /// <summary>
+    ///     Gets the custom variable with the specified <paramref name="name"/> if it exists;
+    ///     otherwise, adds a new one.
+    /// </summary>
+    /// <param name="name">
+    ///     The key of the custom variable to get or add.
+    /// </param>
+    /// <returns>
+    ///     The custom variable corresponding to the specified <paramref name="name"/>.
+    /// </returns>
+    public CustomVariable GetOrAddCustomVariable(string name)
+    {
+        if (!CustomVariables.TryGetValue(name, out CustomVariable variable))
+        {
+            CustomVariables.Add(name, variable = new());
+        }
+
+        return variable;
+    }
+
+    /// <summary>
+    ///     Gets the value of the custom variable corresponding to <paramref name="name"/>.
+    /// </summary>
+    /// <param name="name">
+    ///     The key of the custom variable whose value to get.
+    /// </param>
+    /// <returns>
+    ///     The value of the custom variable corresponding to <paramref name="name"/>.
+    /// </returns>
+    public string CustomVariableValue(string name)
+    {
+        return GetOrAddCustomVariable(name).Value;
+    }
+
+    /// <summary>
+    ///     Sets a custom variable, mapping <paramref name="name"/> to <paramref name="value"/>.
+    /// </summary>
+    /// <param name="name">
+    ///     The key of the custom variable.
+    /// </param>
+    /// <param name="value">
+    ///     The value which the custom variable should have.
+    /// </param>
+    public void SetCustomVariable(string name, string value)
+    {
+        CustomVariable v = GetOrAddCustomVariable(name);
+        v.Value = value;
+        if (v.IsPermanent)
+        {
+            LiveSplitRun.HasChanged = true;
+        }
+    }
+
     public bool UsesEmulator
     {
         get => usesEmulator;
@@ -157,6 +215,7 @@ public class RunMetadata
     {
         LiveSplitRun = run;
         VariableValueNames = new Dictionary<string, string>();
+        CustomVariables = new Dictionary<string, CustomVariable>();
         game = new Lazy<Game>(() => null);
         category = new Lazy<Category>(() => null);
         this.run = new Lazy<SpeedrunComSharp.Run>(() => null);
@@ -324,6 +383,7 @@ public class RunMetadata
             regionName = regionName,
             usesEmulator = usesEmulator,
             VariableValueNames = VariableValueNames.ToDictionary(x => x.Key, x => x.Value),
+            CustomVariables = CustomVariables.ToDictionary(x => x.Key, x => x.Value.Clone()),
             CategoryAvailable = CategoryAvailable,
             GameAvailable = GameAvailable
             //TODO: set members instead later
@@ -334,5 +394,70 @@ public class RunMetadata
     private void TriggerPropertyChanged(bool clearRunID)
     {
         PropertyChanged?.Invoke(this, new MetadataChangedEventArgs(clearRunID));
+    }
+}
+
+/// <summary>
+///     A custom variable that has a value and can be marked permanent.
+/// </summary>
+public sealed class CustomVariable
+{
+    /// <summary>
+    ///     The current value of the custom variable.
+    /// </summary>
+    public string Value { get; set; }
+
+    /// <summary>
+    ///     Gets a value indicating whether the custom variable should be saved in the run permanently.
+    /// </summary>
+    /// <value>
+    ///     <see langword="false"/> initially. Once set to <see langword="true"/>, will remain as such.
+    /// </value>
+    public bool IsPermanent { get; private set; }
+
+    /// <summary>
+    ///     Creates a new instance of the <see cref="CustomVariable"/> class
+    ///     with a default value, not permanent.
+    /// </summary>
+    public CustomVariable()
+        : this(null, false) { }
+
+    /// <summary>
+    ///     Creates a new instance of the <see cref="CustomVariable"/> class
+    ///     with the specified <paramref name="value"/> and permanence.
+    /// </summary>
+    /// <param name="value">
+    ///     The value of the custom variable.
+    /// </param>
+    /// <param name="isPermanent">
+    ///     Indicates whether it should be saved in the run permanently.
+    /// </param>
+    public CustomVariable(string value, bool isPermanent)
+    {
+        Value = value;
+        IsPermanent = isPermanent;
+    }
+
+    /// <summary>
+    ///     Makes the custom variable permanent.
+    /// </summary>
+    /// <returns>
+    ///     The same custom variable object marked permanent.
+    /// </returns>
+    public CustomVariable AsPermanent()
+    {
+        IsPermanent = true;
+        return this;
+    }
+
+    /// <summary>
+    ///     Constructs a new custom variable with the same value and permanence.
+    /// </summary>
+    /// <returns>
+    ///     A copy of the custom variable.
+    /// </returns>
+    public CustomVariable Clone()
+    {
+        return new CustomVariable(Value, IsPermanent);
     }
 }
--- a/src/LiveSplit.Core/Model/RunSavers/XMLRunSaver.cs
+++ b/src/LiveSplit.Core/Model/RunSavers/XMLRunSaver.cs
@@ -45,6 +45,20 @@ public class XMLRunSaver : IRunSaver
         }
 
         metadata.AppendChild(variables);
+
+        XmlElement customVariables = document.CreateElement("CustomVariables");
+        foreach (System.Collections.Generic.KeyValuePair<string, CustomVariable> entry in run.Metadata.CustomVariables)
+        {
+            if (entry.Value.IsPermanent)
+            {
+                XmlElement customVariableElement = ToElement(document, "Variable", entry.Value.Value);
+                customVariableElement.Attributes.Append(ToAttribute(document, "name", entry.Key));
+                customVariables.AppendChild(customVariableElement);
+            }
+        }
+
+        metadata.AppendChild(customVariables);
+
         parent.AppendChild(metadata);
 
         CreateSetting(document, parent, "Offset", run.Offset);


src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -117,6 +117,49 @@ public IDictionary<Variable, VariableValue> VariableValues
             }
         }
 
+        public class CustomVariable
```
just-ero, line 120: Do not nest classes.

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -117,6 +117,49 @@ public IDictionary<Variable, VariableValue> VariableValues
...
+                Value = value;
+            }
+
+            public void ClearValue()
+            {
+                Value = null;
+            }
+        }
+
+        public IDictionary<string, CustomVariable> CustomVariables;
```
just-ero, line 142: What's the benefit of using the interface here?  
What's the reason for making it mutable?  
Do not expose fields publicly.

```suggestion
        public IReadOnlyDictionary<string, CustomVariable> CustomVariables { get; }
```

Or, if truly necessary:
```suggestion
        public IReadOnlyDictionary<string, CustomVariable> CustomVariables { get; private set; }
```

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -117,6 +117,49 @@ public IDictionary<Variable, VariableValue> VariableValues
             }
         }
 
+        public class CustomVariable
+        {
+            public string Value = null;
```
just-ero, line 122: Do not expose fields publicly.

```suggestion
            public string Value { get; set; }
```

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -117,6 +117,49 @@ public IDictionary<Variable, VariableValue> VariableValues
...
+            public bool IsPermanent = false;
```
just-ero, line 123: Does this really need to be mutable?  
Do not expose fields publicly.

```suggestion
            public bool IsPermanent { get; set; }
```

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -117,6 +117,49 @@ public IDictionary<Variable, VariableValue> VariableValues
...
+
+            public CustomVariable Permanent()
+            {
+                IsPermanent = true;
+                return this;
+            }
```
just-ero, line 129: What is the purpose of having this when `IsPermanent` can already be assigned to directly?

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -117,6 +117,49 @@ public IDictionary<Variable, VariableValue> VariableValues
...
+
+            public void SetValue(string value)
+            {
+                Value = value;
+            }
```
just-ero, line 134: What is the purpose of having this when `Value` can already be assigned to directly?

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -117,6 +117,49 @@ public IDictionary<Variable, VariableValue> VariableValues
...
+
+            public void ClearValue()
+            {
+                Value = null;
+            }
```
just-ero, line 139: What is the purpose of having this when `Value` can already be assigned to directly?

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -117,6 +117,49 @@ public IDictionary<Variable, VariableValue> VariableValues
...
+        public CustomVariable CustomVariableMut(string name)
+        {
+            if (CustomVariables.TryGetValue(name, out CustomVariable value))
+            {
+                return value;
+            }
+            else
+            {
+                CustomVariable new_value = new CustomVariable();
+                CustomVariables[name] = new_value;
```
just-ero, line 153:
```suggestion
                CustomVariable newValue = new CustomVariable();
                CustomVariables[name] = newValue;
```

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -117,6 +117,49 @@ public IDictionary<Variable, VariableValue> VariableValues
...
+                return new_value;
```
just-ero, line 154:
```suggestion
                CustomVariable newValue = new CustomVariable();
                CustomVariables[name] = newValue;
                return newValue;
```

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -117,6 +117,49 @@ public IDictionary<Variable, VariableValue> VariableValues
...
+
+            public void ClearValue()
+            {
+                Value = null;
+            }
+        }
+
+        public IDictionary<string, CustomVariable> CustomVariables;
+
+        public CustomVariable CustomVariableMut(string name)
```
just-ero, line 144: What does this name mean? Choose something more descriptive.

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -117,6 +117,49 @@ public IDictionary<Variable, VariableValue> VariableValues
...
+            {
+                return value;
+            }
+            else
+            {
+                CustomVariable new_value = new CustomVariable();
+                CustomVariables[name] = new_value;
+                return new_value;
+            }
+        }
```
just-ero, line 156: What is the purpose of having this when `CustomVariables` is already public?  
What does this method's name mean? Choose something more descriptive.

AlexKnauth: Adding a comment about how it either gets one or creates a new one and mutates the CustomVariables to include it.

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -117,6 +117,37 @@ public IDictionary<Variable, VariableValue> VariableValues
             }
         }
 
+        private Dictionary<string, CustomVariable> CustomVariables;
```
just-ero, line 120:
```suggestion
        private readonly Dictionary<string, CustomVariable> _customVariables;
```

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -117,6 +117,37 @@ public IDictionary<Variable, VariableValue> VariableValues
...
+
+        // Gets a CustomVariable, or creates a new one if it doesn't exist yet,
+        // mutating the CustomVariables to include the new entry
+        private CustomVariable CustomVariableMut(string name)
```
just-ero, line 124:
```suggestion
        private CustomVariable GetOrAddVariable(string name)
```

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -316,4 +349,41 @@ private void TriggerPropertyChanged(bool clearRunID)
...
+
+        public string GetValue()
+        {
+            return Value;
+        }
+
+        public void SetValue(string value)
+        {
+            Value = value;
+        }
```
just-ero, line 373: Turn this into an auto-implemented property.

src/LiveSplit.Core/Model/RunMetadata.cs
AlexKnauth: I tried that, just as I tried using `{ get; }` as in your suggestion above. It broke the `Clone` method below. Making it private should be enough.

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -316,4 +349,41 @@ private void TriggerPropertyChanged(bool clearRunID)
...
+
+        public bool GetIsPermanent()
+        {
+            return IsPermanent;
+        }
```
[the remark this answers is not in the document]
AlexKnauth: It is important that non-permanent variables can be changed to permanent after their creation.

src/LiveSplit.Core/Model/RunMetadata.cs
just-ero: You are violating naming rules that we will be enforcing going forward. Please work around this issue. The field should be readonly, because it is.

AlexKnauth: I can rename it to `_customVariables`, that's fine. But I can't make it readonly or `{ get; }` only without breaking the `Clone` method.

just-ero:
```cs
var clone = new RunMetadata(run)
{
  // ...
};

clone._customVariables = /* ... */;

return clone;
```

just-ero: And I mean seriously, why do we even have this `Clone` method anyway? Is it used anywhere?

src/LiveSplit.Core/Model/RunMetadata.cs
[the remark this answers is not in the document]
just-ero: Can you explain that to me? Why is that the case? I think it should be known upon initialization whether the variable is "permanent" (whatever that means) or not, no?

Otherwise, if this is truly the case, add a `private` setter to the auto-property.

```cs
public bool IsPermanent { get; private set; }
```

src/LiveSplit.Core/Model/RunMetadata.cs
AlexKnauth: Previously, you said `{ get; private set; }` was okay if necessary. I would have thought if `private set` was fine then making the whole thing `private` would also be fine.

I can't just go around deleting `Clone` methods. Even if I searched the `LiveSplit` organization and didn't find any uses, that wouldn't mean there are no uses in components written by others.

just-ero: Fields must be private, properties must be public. It's as simple as that. You either expose the entire collection publicly via `public Dictionary<string, CustomVariable> CustomVariables { get; private set; }` (although this will undoubtedly allow people to directly manipulate the dictionary), or you use a field and do not expose it: `private Dictionary<string, CustomVariable> _customVariables;`.  
Since the dictionary itself (the reference) never changes, it would not make sense to not make the field `readonly`.

just-ero: I'm willing to backpedal on "properties must be public".

Maybe we can settle for `private Dictionary<string, CustomVariable> CustomVariables  { get; set; } = new();`.

src/LiveSplit.Core/Model/RunMetadata.cs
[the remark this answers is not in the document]
AlexKnauth: When an autosplitter adds a Custom Variable, it is not permanent. But when the user themselves adds a Custom Variable, it is permanent. A future UI implementation could give the user a list of Custom Variables which may not be permanent yet, and the user could select one and make it permanent.

I haven't implemented that UI because my use case doesn't need user-added/permanent custom variables. But the corresponding Custom Variables in LiveSplit One do allow this, so I wanted to leave room for it.

src/LiveSplit.Core/Model/RunMetadata.cs
AlexKnauth: Changing `CustomVariables` to a private property.

src/LiveSplit.Core/Model/RunMetadata.cs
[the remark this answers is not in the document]
just-ero: Alright. Perhaps name the method which sets this `MakePermanent` or something.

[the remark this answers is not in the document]
AlexKnauth: Done.

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -316,4 +349,26 @@ private void TriggerPropertyChanged(bool clearRunID)
             PropertyChanged?.Invoke(this, new MetadataChangedEventArgs(clearRunID));
         }
     }
+
+    internal class CustomVariable
+    {
+        public string Value { get; set; } = null;
+        // monotonic: once true, never set back to false
+        public bool IsPermanent { get; private set; } = false;
```
just-ero, line 357: Assignment of default value is unnecessary.

```suggestion
        public string Value { get; set; }
        // monotonic: once true, never set back to false
        public bool IsPermanent { get; private set; }
```

just-ero:
After this, I would perhaps ask you to add short documentation comments (`///`) to all public APIs. That would be the following:
```
public event EventHandler CustomVariableSet
public void SetCustomVariable(string name, string value)
public string CustomVariableValue(string name)
```

`CustomVariable` is `internal` and therefore does not necessarily need documentation.

just-ero:
Also thanks a lot for your cooperation on my review! I hope I didn't overdo it. I would like LiveSplit's codebase to move towards more consistency and I suggested what I thought was necessary for that.

AlexKnauth:
I added doc comments to those 3 public things, and I converted what comments there were on private things into doc comments as well.

src/LiveSplit.Core/Model/LiveSplitState.cs
```diff
@@ -91,6 +91,12 @@ public bool IsGameTimePaused
         public event EventHandler RunManuallyModified;
         public event EventHandler ComparisonRenamed;
 
+        /// <summary>
+        /// Event <c>CustomVariableSet</c> is invoked when setting a custom variable.
+        /// </summary>
+        /// <seealso cref="SetCustomVariable"/>
```
just-ero, line 97:
```suggestion
        /// <summary>
        ///     Invoked when setting a custom variable.
        /// </summary>
```

src/LiveSplit.Core/Model/LiveSplitState.cs
```diff
@@ -213,6 +219,22 @@ public void SetGameTime(TimeSpan? gameTime)
             }
         }
 
+        /// <summary>
+        /// Method <c>SetCustomVariable</c> sets a custom variable mapping <c>name</c> to <c>value</c>.
+        /// Invokes the event <c>CustomVariableSet</c> to alert interested listeners.
+        /// </summary>
+        /// <param name="name">The name can be used as a key to look up the custom variable.</param>
+        /// <param name="value">The value to set the custom variable to.</param>
+        /// <seealso cref="CustomVariableSet"/>
```
just-ero, line 228:
```suggestion
        /// <summary>
        ///     Sets a custom variable, mapping <paramref name="name"/> to <paramref name="value"/>.
        /// </summary>
        /// <param name="name">
        ///     The key of the custom variable.
        /// </param>
        /// <param name="value">
        ///     The value which the custom variable should have.
        /// </param>
        /// <remarks>
        ///     Invokes <see cref="CustomVariableSet"/>.
        /// </remarks>
```

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -117,6 +117,54 @@ public IDictionary<Variable, VariableValue> VariableValues
...
         }
 
+        private Dictionary<string, CustomVariable> CustomVariables { get; set; } = new Dictionary<string, CustomVariable>();
+
+        /// <summary>
+        /// Method <c>GetOrAddCustomVariable</c> either gets a CustomVariable if it exists,
+        /// or creates a new one otherwise, mutating the CustomVariables to include it.
+        /// </summary>
+        /// <param name="name">The name of the custom variable to look up or create.</param>
+        /// <returns>The new or existing <c>CustomVariable</c> object for <c>name</c>.</returns>
```
just-ero, line 127:
```suggestion
        /// <summary>
        ///     Gets a the custom variable with the specified <paramref name="name"/> if it exists;
        ///     otherwise, adds a new one.
        /// </summary>
        /// <param name="name">
        ///     The key of the custom variable to get or add.
        /// </param>
        /// <returns>
        ///     The custom variable corresponding to the specified <paramref name="name"/>.
        /// </returns>
```

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -117,6 +117,54 @@ public IDictionary<Variable, VariableValue> VariableValues
...
+                CustomVariables[name] = newValue;
+                return newValue;
+            }
+        }
+
+        /// <summary>
+        /// Method <c>CustomVariableValue</c> gets the value of the custom variable for <c>name</c>.
+        /// </summary>
+        /// <param name="name">The name of the custom variable to look up.</param>
+        /// <returns>The value of the custom variable for <c>name</c>.</returns>
```
just-ero, line 146:
```suggestion
        /// <summary>
        ///     Gets the value of the custom variable corresponding to <paramref name="name"/>.
        /// </summary>
        /// <param name="name">
        ///     The key of the custom variable whose value to get.
        /// </param>
        /// <returns>
        ///     The value of the custom variable corresponding to <paramref name="name"/>.
        /// </returns>
```

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -117,6 +117,54 @@ public IDictionary<Variable, VariableValue> VariableValues
...
+
+        /// <summary>
+        /// Method <c>SetCustomVariable</c> sets a custom variable mapping <c>name</c> to <c>value</c>.
+        /// Unlike the method on <c>LiveSplitState</c> of the same name,
+        /// this does not invoke the the event <c>CustomVariableSet</c>,
+        /// so callers should make sure the event is invoked separately.
+        /// </summary>
+        /// <param name="name">The name can be used as a key to look up the custom variable.</param>
+        /// <param name="value">The value to set the custom variable to.</param>
+        /// <returns>True when the custom variable is permanent, false otherwise.</returns>
```
just-ero, line 160:
```suggestion
        /// <summary>
        ///     Sets a custom variable, mapping <paramref name="name"/> to <paramref name="value"/>.
        /// </summary>
        /// <param name="name">
        ///     The key of the custom variable.
        /// </param>
        /// <param name="value">
        ///     The value which the custom variable should have.
        /// </param>
        /// <returns>
        ///     A value indicating whether the custom variable is permanent.
        /// </returns>
        /// <remarks>
        ///     Does not invoke <see cref="LiveSplitState.CustomVariableSet"/>. Make sure to invoke the event separately.
        /// </remarks>
```

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -316,4 +366,30 @@ private void TriggerPropertyChanged(bool clearRunID)
...
     }
+
+    internal class CustomVariable
+    {
+        public string Value { get; set; }
+        /// <summary>
+        /// Property <c>IsPermanent</c> is true when the custom variable should be saved
+        /// permanently when a run is saved to a file, false otherwise.
+        /// Once true, it should never be set back to false.
+        /// </summary>
```
just-ero, line 377:
```suggestion

        /// <summary>
        ///     Gets a value indicating whether the custom variable should be saved in the run permanently.
        /// </summary>
        /// <value>
        ///     <see langword="false"/> initially. Once set to <see langword="true"/>, will remain as such.
        /// </value>
```

just-ero:
Please make sure these comments are actually valid and link to the types accordingly, I have not tested them.

just-ero:
I want to take this opportunity to ask about the design as a whole.

Why does `LiveSplitState` need to be involved here? It should not really handle anything pertaining to these custom variables and it leads to some duplicate code.

As I look at this more, I can also see that getting a variable should perhaps not return its value, but an instance of `CustomVariable` instead. The user can then access `Value` and `IsPermanent` separately. This is a major point I want you to address.

Furthermore, I'm not entirely sure the run needs to know about variables at all? The AutoSplittingRuntime component alone should be the one storing this data, in my opinion. It can exist without being tied to a run, after all (in the layout).

Finally, I'm not really sure whether it's appropriate to return the `IsPermanent` state in `internal bool SetCustomVariable`. As with the above suggestion, it could simply return the created instance of `CustomVariable`.

AlexKnauth:
My reason for putting them on `LiveSplitState` is for the event `CustomVariableSet`. So many events like `OnStart, `OnSplit`, `OnReset`, etc. are defined there, it seemed like the place to go for the new event I needed. Then the public method `SetCustomVariable` goes in the same place so it can invoke the event.

If there's a better way to structure the event, I can move it somewhere else as long as it's still easily accessible from the `LiveSplitState` that a component would have access to.

Thinking again about returning the `IsPermanent` state in `internal bool SetCustomVariable`, I think I agree. I can remove `internal bool SetCustomVariable` entirely if I make both the `CustomVariable` class and the `GetOrAddCustomVariable` method public. Future work on the UI to allow users to make permanent custom variables will require them to be public anyway, even if current work does not.

just-ero:
The `OnStart`, `OnSplit`, `OnReset` events are also misplaced and should have gone into a class that has something to do with the timer, like `TimerModel`. This event could go directly into `RunMetadata`, to which the user has access to via `LiveSplitState.Run.Metadata`.

Sorry to bother about this again, but perhaps the name `AsPermanent` would be a better fit for `MakePermanent`. It somewhat implies that it returns the instance of the `CustomVariable` *as* one that is permanent.

What do you think of the design I proposed, exposing `CustomVariable` publicly and just passing around instances of that?

Another thing I wanted to ask is, are "custom variables" always just strings? Never any other types?

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -117,6 +117,46 @@ public IDictionary<Variable, VariableValue> VariableValues
             }
         }
 
+        private Dictionary<string, CustomVariable> CustomVariables { get; set; } = new Dictionary<string, CustomVariable>();
+
+        /// <summary>
+        ///     Gets a the custom variable with the specified <paramref name="name"/> if it exists;
```
just-ero, line 123:
```suggestion
        ///     Gets the custom variable with the specified <paramref name="name"/> if it exists;
```

just-ero:
What I would like to put on the table is an API like the following:
```cs
public sealed class CustomVariableMap : IReadOnlyDictionary<string, CustomVariable>
{
    private readonly Dictionary<string, CustomVariable> _map = new Dictionary<string, CustomVariable>();

    public CustomVariable this[string key]
    {
        get
        {
            if (!_map.TryGetValue(key, out var value))
            {
                _map.Add(key, value = new CustomVariable());
            }

            return value;
        }
    }

    public int Count => _map.Count;

    public IEnumerable<string> Keys => _map.Keys;
    public IEnumerable<CustomVariable> Values => _map.Values;

    public bool ContainsKey(string key)
    {
        return _map.ContainsKey(key);
    }

    public bool TryGetValue(string key, [MaybeNullWhen(false)] out CustomVariable value)
    {
        return _map.TryGetValue(key, out value);
    }

    public IEnumerator<KeyValuePair<string, CustomVariable>> GetEnumerator()
    {
        return _map.GetEnumerator();
    }

    IEnumerator IEnumerable.GetEnumerator()
    {
        return GetEnumerator();
    }
}
```
```cs
public class RunMetadata
{
  public CustomVariableMap Variables { get; } = new CustomVariableMap();
}
```

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -316,4 +358,46 @@ private void TriggerPropertyChanged(bool clearRunID)
             PropertyChanged?.Invoke(this, new MetadataChangedEventArgs(clearRunID));
         }
     }
+
+    /// <summary>
+    ///     A custom variable that has a value and can be marked permanent.
+    /// </summary>
+    public class CustomVariable
```
just-ero, line 365:
```suggestion
    public sealed class CustomVariable
```

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -316,4 +358,46 @@ private void TriggerPropertyChanged(bool clearRunID)
...
+        /// <value>
+        ///     <see langword="false"/> initially. Once set to <see langword="true"/>, will remain as such.
+        /// </value>
+        public bool IsPermanent { get; private set; }
+        
+        /// <summary>
+        ///     Makes the custom variable permanent.
+        /// </summary>
+        /// <returns>The same custom variable object marked permanent.</returns>
+        public CustomVariable MakePermanent()
```
just-ero, line 384:
```suggestion
        public CustomVariable AsPermanent()
```

just-ero:
I think I might like to see these constructors added to `CustomVariable` as well:
```cs
public CustomVariable()
    : this(false) { }

public CustomVariable(bool isPermanent)
{
    IsPermanent = isPermanent;
}
```

AlexKnauth:
I don't expect anyone to construct permanent `CustomVariable` instances directly with `CustomVariable(isPermanent)`. Instead I expect them to use `GetOrAddCustomVariable(name).AsPermanent(true)`, so that if it already exists as a non-permanent custom variable, it doesn't create a new one.

> Are these custom variables saved in any way? Are they loaded? Where do they come from?

Currently there is no code that would call `AsPermanent` to create a custom variable, but once there is, custom variables would come from 2 places:
- 1. The autosplitter, triggering calls to `SetCustomVariable(name, value)` to create non-permanent custom variables and set values of existing ones (without changing their permanence).
- 2. The user through a not-yet-written UI which I expect to call `GetOrAddCustomVariable(name).AsPermanent(true)` to create permanent custom variables and make existing ones permanent.

Since there is no code that needs to create permanent custom variables, I haven't written the code  in the RunSavers to save the permanent ones yet. The use-case I need custom variables for, only needs (1) the temporary ones from the autosplitter. Should I write that code in the RunSavers, or would it be better to remove `IsPermanent` and make them all temporary, since that's all I need for now?

just-ero:
If the custom variables are confined to the autosplitter for now, why is `RunMetadata` involved? Please address the other questions.

I don't see a problem in constructing `CustomVariable`. If the user wants, they could then store it in `CustomVariables` (if it were exposed). What you designed here is a bit unconventional.

wooferzfg:
Converting to draft since there's open questions

AlexKnauth:
I've added parsing and saving custom variables in the `Metadata` section of `.lss` splits files. Custom variables from LiveSplit One can be parsed as custom variables in LiveSplit this way. And in the other direction, custom variables modified by LiveSplit can be imported into LiveSplit One and displayed there.

That location in `.lss` splits files produced by LiveSplit One is the reason it's on the RunMetadata here.

just-ero:
> How is it handled when there are multiple instances of the AutoSplittingRuntime component present? One on the run, and one or more on the layout?

> Do the AutoSplittingRuntime components on the layout all add their variables to the current Run?

> How do I distinguish where the event came from?

> Perhaps it would make sense to simply drop the event and the way to set any variables from the state for now and add the event again later when there's demand for it?

src/LiveSplit.Core/Model/RunFactories/StandardFormatsRunFactory.cs
```diff
@@ -134,6 +134,15 @@ public IRun Create(IComparisonGeneratorsFactory factory)
...
         }
 
+        using (LiveSplitCore.RunMetadataCustomVariablesIter iter = metadata.CustomVariables())
+        {
+            LiveSplitCore.RunMetadataCustomVariableRef variable;
+            while ((variable = iter.Next()) != null)
+            {
+                run.Metadata.GetOrAddCustomVariable(variable.Name()).AsPermanent().Value = variable.Value();
+            }
+        }
```
just-ero, line 144: Use simple `using` statements.

```suggestion
        using var variables = metadata.CustomVariables();
        
        while (variables.Next() is { } variable)
        {
            run.Metadata.GetOrAddCustomVariable(variable.Name()).AsPermanent().Value = variable.Value();
        }
```

I don't know where this stuff comes from, but why are `variable.Name()` and `variable.Value()` methods?

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -119,6 +119,49 @@ public IDictionary<Variable, VariableValue> VariableValues
...
+        {
+            return value;
+        }
+        else
+        {
+            CustomVariable newValue = new CustomVariable();
+            CustomVariables[name] = newValue;
+            return newValue;
+        }
+    }
```
just-ero, line 149:
```suggestion
    public CustomVariable GetOrAddCustomVariable(string name)
    {
        if (!CustomVariables.TryGetValue(name, out CustomVariable variable))
        {
            CustomVariables.Add(variable = new());
        }

        return variable;
    }
```

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -336,3 +381,63 @@ private void TriggerPropertyChanged(bool clearRunID)
...
+    /// <summary>
+    ///     Gets a value indicating whether the custom variable should be saved in the run permanently.
+    /// </summary>
+    /// <value>
+    ///     <see langword="false"/> initially. Once set to <see langword="true"/>, will remain as such.
+    /// </value>
+    public bool IsPermanent { get; private set; }
+
+    /// <summary>
+    ///     Constructs an empty temporary custom variable.
```
just-ero, line 404:
```suggestion
    ///     Creates a new instance of the <see cref="CustomVariable"/> class
    ///     with a default value and temporary permanence.
```

Not big on the phrasing "temporary permanence", maybe you can think of something better.

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -336,3 +381,63 @@ private void TriggerPropertyChanged(bool clearRunID)
...
+    /// </summary>
+    public CustomVariable()
+        : this(null, false) { }
+
+    /// <summary>
+    ///     Constructs a custom variable with the given <paramref name="value"/>,
+    ///     as permanent if <paramref name="value"/> is <see langword="true"/>.
```
just-ero, line 411:
```suggestion
    ///     Creates a new instance of the <see cref="CustomVariable"/> class
    ///     with the specified <paramref name="value"/> and permanence.
```

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -336,3 +381,63 @@ private void TriggerPropertyChanged(bool clearRunID)
...
+    public CustomVariable(string value, bool isPermanent)
+    {
+        Value = value;
+        IsPermanent = isPermanent;
+    }
+
+    /// <summary>
+    ///     Makes the custom variable permanent.
+    /// </summary>
+    /// <returns>The same custom variable object marked permanent.</returns>
```
just-ero, line 428:
```suggestion
    /// <returns>
    ///     The same custom variable object marked permanent.
    /// </returns>
```

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -336,3 +381,63 @@ private void TriggerPropertyChanged(bool clearRunID)
...
+    public CustomVariable AsPermanent()
+    {
+        IsPermanent = true;
+        return this;
+    }
+
+    /// <summary>
+    ///     Constructs a new custom variable with the same value and permanence.
+    /// </summary>
+    /// <returns>A copy of the custom variable.</returns>
```
just-ero, line 438:
```suggestion
    /// <returns>
    ///     A copy of the custom variable.
    /// </returns>
```

AlexKnauth:
> How is it handled when there are multiple instances of the AutoSplittingRuntime component present? One on the run, and one or more on the layout?

Context:

When there are multiple autosplitters, they can "fight" over the GameTime. I've seen this happen before: when one autosplitter wants the GameTime running and another wants the GameTime paused, the user sees the GameTime running more slowly than usual. On each tick, whichever autosplitter sent its command last, wins that tick, so the timer is paused some ticks and unpaused other ticks. This is bad, so I tell users to disable the default autosplitter on the run when they use my autosplitter via the layout.

Answer:

When there are multiple autosplitters, they can both set custom variables and theoretically "fight" over a custom variable they have in common. However, it's much less likely to cause problems than the GameTime fighting above. Custom variables are only for output information, not for storing state: an ASR autosplitter cannot read them later, only write, and so it cannot be caught off-guard by them being changed by another autosplitter.

There are 2 ways LiveSplit One can display custom variables to the user: in a Text component that the user sets to display one, or in a column in the Splits component that the user sets to display one. Those can be implemented in LiveSplit after these Core changes are settled.

In the case of a "fight" over a custom variable when multiple autosplitters are each trying to set the same variable to different values every tick, the user might see the text flickering between the different values, both on the Text component and on the column in the current split of the Splits component. On the previous splits in the Splits component, the column will show the single value of that custom variable from the instant that the split ended, so an autosplitter can set the custom variable just before it autosplits, to ensure it's more likely to win and display that column information on the split it's responsible for ending.

> Do the AutoSplittingRuntime components on the layout all add their variables to the current Run?

Yes.

> How do I distinguish where the event came from?

The event itself doesn't carry this information. When a component wants to display a specific custom variable, it can read the value of the custom variable it cares about after receiving the event, and see whether it actually changed.

> Perhaps it would make sense to simply drop the event and the way to set any variables from the state for now and add the event again later when there's demand for it?

That would require a component that wants to display a custom variable to read the custom variable it cares about every tick to see if it changed, rather than only on the ticks where some custom variable changed. That's probably do-able, just less convenient than what LiveSplit One does, which has an event for this.

src/LiveSplit.Core/Model/RunFactories/StandardFormatsRunFactory.cs
AlexKnauth: I'm using the same style for this `CustomVariables` iteration as was already used for the `SpeedrunComVariables` iteration just above it. Since there are two such iterations, I fear that flattening it out like you propose will make it harder to read, not easier.

The `.Name()` and `.Value()` methods come from the `RunMetadataCustomVariableRef` class in the generated `LiveSplitCore.g.cs` file, a step on the bridge between Rust code and C Sharp code.

just-ero:
Other than the fact that I don't really know why `LiveSplitState` needs to be involved here, it looks good now.

AlexKnauth:
I can delete the event, I've tested out a component to not require the event anymore.

just-ero:
I just find it misplaced in `LiveSplitState`, no?

AlexKnauth:
Is the `LiveSplitRun` found in `RunMetadata` guaranteed to be the same run as the `Run` found in `LiveSplitState`, given that the `Run` in `LiveSplitState` contains the `RunMetadata` instance?

just-ero:
That's beyond me. I find the entire architecture around this ridiculous. But we're in too deep at this point.  
@wooferzfg @CryZe Any idea?

AlexKnauth:
I have removed the event, and moved `SetCustomVariable` from `LiveSplitState` to `RunMetadata`.

AlexKnauth:
Bump, can this be merged soon?

PR #2531, opened 2024-08-30
Clean up code

10-days-till-xmas:


src/LiveSplit.Core/TimeFormatters/ShortTimeFormatter.cs
```diff
@@ -30,10 +30,5 @@ public string Format(TimeSpan? time, DigitsFormat format)
 
             return formatRequest.Format(time);
         }
-
-        public string Format(TimeSpan? timeSpan)
-        {
-            return Format(timeSpan, DigitsFormat.SingleDigitSeconds);
-        }
```
wooferzfg, line 37: We should keep this in case components (maybe third party components) are using it

10-days-till-xmas: Shouldn't it be fine though since the changes made to the other Format() would cover it too in the exact same way. It'd likely be compiled in the same way too

wooferzfg: Are you sure it compiles exactly the same way? Can you do a hash comparison?

src/LiveSplit.Core/TimeFormatters/ShortTimeFormatter.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace LiveSplit.TimeFormatters
 {
-    public class ShortTimeFormatter : ITimeFormatter
+    public class ShortTimeFormatter : GeneralTimeFormatter
```
wooferzfg, line 6: Why inherit from `GeneralTimeFormatter` when `Format` doesn't use it?

src/LiveSplit.Core/TimeFormatters/ShortTimeFormatter.cs
wooferzfg: I find it a bit hard to believe since you're changing which class it inherits from

src/LiveSplit.Core/TimeFormatters/ShortTimeFormatter.cs
10-days-till-xmas: It led to build errors if I didn't change this. And it shouldn't change anything either way wouldn't it?

wooferzfg: Why are there build errors?

10-days-till-xmas: In abt 30 mins I'll be back at my computer to send a screenshot of the errors and do that hash comparison

10-days-till-xmas: In abt 30 mins I'll be back at my computer to send a screenshot of the errors and do that hash comparison

wooferzfg: My guess is that it's because you deleted the function that you need to implement https://github.com/LiveSplit/LiveSplit/blob/master/src/LiveSplit.Core/TimeFormatters/ITimeFormatter.cs#L7

wooferzfg: I would suggest just dropping the `ShortTimeFormatter.cs` changes from this PR

src/LiveSplit.Core/TimeFormatters/ShortTimeFormatter.cs
just-ero: woofer is correct, this is a binary breaking change. Assemblies currently calling this overload will break, because the signature no longer points to an existing method.

src/LiveSplit.Core/TimeFormatters/ShortTimeFormatter.cs
just-ero: > My guess is that it's because you deleted the function that you need to implement

That wouldn't be it. `GeneralTimeFormatter` implements that method.

src/LiveSplit.Core/TimeFormatters/ShortTimeFormatter.cs
just-ero: Actually, I retract that. `GeneralTimeFormatter` contains the `string Format(TimeSpan?)` signature.  
However, I'm still against this change. I don't think `ShortTimeFormatter` needs to know about `GeneralTimeFormatter`.

Relevant docs: https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md#members (second bullet).

PR #2533, opened 2024-09-02
Add `.editorconfig` & apply style rules

just-ero:
### Description

* Introduces an `.editorconfig` file to enforce consistent style rules across LiveSplit projects.
* Upgrades the language version to C# 12 (most recent stable).
* Adds `PolySharp` to all projects to allow for support of some language features which require runtime support (`Index` & `Range` operators).
* Fixes some warnings.


wooferzfg:
Why did the deep pointer tests need to change?

wooferzfg:
Can we exclude `LiveSplitCore.cs`?

Screenshot_20240902_103834_Chrome Beta

wooferzfg:
Why are there still some build warnings?

just-ero:
`DeepPointer.InitializeOffsets` now uses a collection initializer instead of `List<T>.AddRange`. Passing a `null` collection to `AddRange` causes `ArgumentNullException`. Passing a `null` collection into a collection initializer causes `NullReferenceException`.

just-ero:
We cannot exclude `LiveSplitCore.cs` this way.

wooferzfg:
If we're not fixing them, can we suppress the warnings in these particular spots so that they don't show up in the CI build?

wooferzfg:
> We cannot exclude `LiveSplitCore.cs` this way.

Maybe filter the warnings in GitHub Actions then? We need some way to keep the build output clean

just-ero:
We can rename the file from `LiveSplitCore.cs` to `LiveSplitCore.g.cs`. This also indicates that the code is generated and excludes it from analysis.

wooferzfg:
> We can rename the file from `LiveSplitCore.cs` to `LiveSplitCore.g.cs`. This also indicates that the code is generated and excludes it from analysis.

That works

PR #2547, opened 2024-11-11
Change the response type of CommandServer to normalize the response.

Someone999:
Sometimes, I need to match each request and response, so I added a response class to normalize the response.
Response format
&gt;&gt; //Start mark
RequestId: xxx //User-defined id
Success: True //True or False
State: //May be empty, Failed, or Exception at current
Command: xxx // Original command text
myData:  myValue //Custom data
<<// End Mark

e.g.

&gt;&gt; 
RequestId: 0
Success: True
State: 
Command: getcurrentsplitindex
index: 1
<<

or

&gt;&gt; 
RequestId: 1
Success: false
State: Failed 
Command: setgametime
message:  Failed to parse time while setting game time: 0
<<

Request format
RquestId|CommandText
e.g.
0|startorsplit
1|setgametime 5


wooferzfg:
Does this break existing applications that use the server?

Someone999:
You need I can add a fallback.

Someone999:
Because the start mark, previous application can not parse this.

PR #2551, opened 2024-12-11
Add option to make the background transparent

RibShark:
This adds an option to make the background of LiveSplit transparent + blurred. This allows for the background to appear transparent without the opacity of the text being reduced alongside it. OBS can capture this (without blur) with the Windows 10 window capture method.


src/LiveSplit.View/View/LayoutSettingsControl.cs
```diff
@@ -1,187 +1,188 @@
...
-        lblOutlines.Enabled = btnTextOutlineColor.Enabled = chkAntiAliasing.Checked;
-    }
-
-    private void LayoutSettingsControl_Load(object sender, EventArgs e)
-    {
-        chkRainbow_CheckedChanged(null, null);
-        chkAntiAliasing_CheckedChanged(null, null);
-    }
-}
+﻿using System;
```
wooferzfg, line 1: Same thing for this file, please clean up the diff

wooferzfg:
How does this new setting interact with setting the background color, background image, or opacity?

src/LiveSplit.View/View/LayoutSettingsControl.Designer.cs
```diff
@@ -28,866 +28,937 @@ protected override void Dispose(bool disposing)
...
+            this.chkTransparentBackground.Location = new System.Drawing.Point(3, 80);
+            this.chkTransparentBackground.Name = "chkTransparentBackground";
+            this.chkTransparentBackground.Size = new System.Drawing.Size(595, 20);
+            this.chkTransparentBackground.TabIndex = 23;
+            this.chkTransparentBackground.Text = "Transparent Background";
+            this.chkTransparentBackground.UseVisualStyleBackColor = true;
+            // 
+            // LayoutSettingsControl
+            // 
+            this.AutoScaleDimensions = new System.Drawing.SizeF(8F, 16F);
```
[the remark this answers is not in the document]
RibShark: Seems visual studio did that automatically when I was editing the form, I will see if removing that change affects anything.

src/LiveSplit.Core/UI/LayoutSavers/XMLLayoutSaver.cs
```diff
@@ -1,126 +1,127 @@
...
-        parent.Attributes.Append(SettingsHelper.ToAttribute(document, "version", "1.6.1"));
-        CreateLayoutNode(document, parent, layout);
-        document.AppendChild(parent);
-
-        document.Save(stream);
-    }
-}
-
+﻿using System;
+using System.Collections.Generic;
```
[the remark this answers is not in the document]
RibShark: Oops, sorry for this (and the other one), I'll get onto fixing that in a bit.

RibShark:
Here's a [video](https://www.youtube.com/watch?v=9gRIa4rCbBI) demonstrating what it looks like. Works fine for the most part, though I just noticed doesn't seem to have an effect with images when the blur setting is off, will take a look and see what I can do to fix this.

src/LiveSplit.View/View/LayoutSettingsControl.Designer.cs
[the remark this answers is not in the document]
wooferzfg: It could be because your scaling in the editor isn't set to 100%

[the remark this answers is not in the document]
wooferzfg: Please configure your editor in VS to avoid unnecessary changes rather than trying to edit this file manually

src/LiveSplit.Core/UI/LayoutSavers/XMLLayoutSaver.cs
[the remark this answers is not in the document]
RibShark: Done!

src/LiveSplit.View/View/LayoutSettingsControl.cs
RibShark: Done!

wooferzfg:
Do you need to check if the transparency setting changed here? https://github.com/LiveSplit/LiveSplit/blob/4e6afeaa34330474f61b79be69d5d7094e619068/src/LiveSplit.View/View/TimerForm.cs#L1558

RibShark:
> Do you need to check if the transparency setting changed here?
> 
> https://github.com/LiveSplit/LiveSplit/blob/4e6afeaa34330474f61b79be69d5d7094e619068/src/LiveSplit.View/View/TimerForm.cs#L1556-L1559

No, it seems to work without that, all that matters is that the background is a Bitmap, that conversion only needs to happen when the image is initially loaded.

src/LiveSplit.View/View/TimerForm.cs
```diff
@@ -66,6 +66,25 @@ public partial class TimerForm : Form
...
+            {
+                TransparentBackgroundState = value;
+                var attributeData = new WindowCompositionAttributeData()
+                {
+                    Attribute = WCA_ACCENT_POLICY,
+                    Data = value ? AccentPolicyBlurPtr : AccentPolicyDisabledPtr,
+                    SizeOfData = Marshal.SizeOf(typeof(AccentPolicy))
+                };
+                SetWindowCompositionAttribute(Handle, attributeData);
+            }
```
just-ero, line 84: Use unsafe context and do the following:

```cs
if (value == _transparentBackground)
{
    return;
}

AccentPolicy policy = new()
{
    AccentState = value ? ACCENT_ENABLE_BLURBEHIND : ACCENT_DISABLED
};

WindowCompositionAttributeData data = new()
{
    Attribute = WCA_ACCENT_POLICY,
    Data = &policy,
    SizeOfData = sizeof(AccentPolicy)
};

SetWindowCompositionAttribute(Handle, data);
_transparentBackground = value;
```

src/LiveSplit.View/View/TimerForm.cs
```diff
@@ -171,6 +190,42 @@ private bool MousePassThrough
...
+    {
+        AccentState = ACCENT_ENABLE_BLURBEHIND,
+    };
+    private static readonly IntPtr AccentPolicyBlurPtr = Marshal.AllocHGlobal(Marshal.SizeOf(typeof(AccentPolicy)));
+
+    private AccentPolicy AccentPolicyDisabled = new()
+    {
+        AccentState = ACCENT_DISABLED,
+    };
+    private static readonly IntPtr AccentPolicyDisabledPtr = Marshal.AllocHGlobal(Marshal.SizeOf(typeof(AccentPolicy)));
```
just-ero, line 224: Remove all of this.

src/LiveSplit.View/View/TimerForm.cs
```diff
@@ -66,6 +66,25 @@ public partial class TimerForm : Form
...
+                {
+                    Attribute = WCA_ACCENT_POLICY,
+                    Data = value ? AccentPolicyBlurPtr : AccentPolicyDisabledPtr,
+                    SizeOfData = Marshal.SizeOf(typeof(AccentPolicy))
+                };
+                SetWindowCompositionAttribute(Handle, attributeData);
+            }
+        }
+    }
+    private bool TransparentBackgroundState = false;
```
just-ero, line 87:
```suggestion
    private bool _transparentBackground;
```

src/LiveSplit.View/View/TimerForm.cs
just-ero, line 1: This file shouldn't really be cluttered up with Win32 definitions, those should go into separate files.

src/LiveSplit.View/View/TimerForm.cs
```diff
@@ -171,6 +190,42 @@ private bool MousePassThrough
...
+    private static readonly IntPtr AccentPolicyBlurPtr = Marshal.AllocHGlobal(Marshal.SizeOf(typeof(AccentPolicy)));
+
+    private AccentPolicy AccentPolicyDisabled = new()
+    {
+        AccentState = ACCENT_DISABLED,
+    };
+    private static readonly IntPtr AccentPolicyDisabledPtr = Marshal.AllocHGlobal(Marshal.SizeOf(typeof(AccentPolicy)));
+
+    [DllImport("user32.dll")]
+    private static extern int SetWindowCompositionAttribute(IntPtr hWnd, in WindowCompositionAttributeData data);
```
just-ero, line 227:
```suggestion
    private static extern int SetWindowCompositionAttribute(nint hWnd, WindowCompositionAttributeData* data);

```

wooferzfg:
> No, it seems to work without that, all that matters is that the background is a Bitmap, that conversion only needs to happen when the image is initially loaded.

What if the conversion doesn't happen when you initially create the baked background image, and then you enable the transparency setting?

src/LiveSplit.View/View/TimerForm.cs
RibShark: I wrote it like that to be consistent with MousePassThroughState later on in the file, which is used in a similar pattern. I can change the name, but it will be less consistent that way.

src/LiveSplit.View/View/TimerForm.cs
RibShark: Did this because there were already a lot of Win32 definitions in TimerForm. I'll move the definitions to LiveSplit.View/Model/Win32.cs unless you have a better suggestion.

RibShark:
> > No, it seems to work without that, all that matters is that the background is a Bitmap, that conversion only needs to happen when the image is initially loaded.
> 
> What if the conversion to Bitmap doesn't happen when you initially create the baked background image (because transparency is initially off), and then you enable the transparency setting? (Did you test every combination of settings changing?)

Oops, forgot about that possibility, will add the check.

ImAciidz:
> OBS can capture this (without blur) with the Windows 10 window capture method

On Windows 10 (tested 22H2, OBS 31), the blur is still captured:
win10

On Windows 11 (tested 23H2 & 24H2, OBS 31), there's a noise of sorts in the capture, which results in a gray background:
win11

So unless I'm misunderstanding the purpose of this (or doing something wrong), it unfortunately doesn't seem very useful.

Skajdrowski:
@ImAciidz The gray noise you're talking about is barely visible [on a game that covers entire screen](https://www.youtube.com/watch?v=yNxK1nfWjGI)
It's a very nice feature for fonts with round edges. They don't get butchered by OBS Chroma keying anymore.

ImAciidz:
It's worth noting that the feature doesn't work at all if transparency effects are disabled in the Windows accessibility settings.

PR #2565, opened 2025-01-27
CommandServer: add getcustomvariablevalue

### LINKED ISSUE #2564: make CustomVariables accessible through server component
Adding a command to the (websocket) server to let us retrieve custom variables would let you use livesplit for much more data gathering and stream fx than just a timer.

Id personally like to use it to have my vtuber die when my game character dies among other things.

AlexKnauth:
Resolves #2564 by adding `getcustomvariablevalue` to let a client retrieve a custom variable value.

I have not tested this yet, as I don't have a client to test it with. @Luukie13, do you have such a client?

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,7 @@ Commands that return a string:
 - getcurrentsplitname  
 - getprevioussplitname
 - getcurrenttimerphase
+- getcustomvariablevalue NAME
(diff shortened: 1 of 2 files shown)


wooferzfg:
The logic looks correct to me

PR #2572, opened 2025-02-12
Set DPI Awareness at Runtime

### LINKED ISSUE #2563: DPI Awareness Option
The new DPI awareness in 1.8.30 completely screwed my layout and in my opinion made live split look WAY WORSE because of how much smaller it was. Please add OPTIONS to allow someone to choose what they want when deciding to completely change something out of the blue. I immediately had to downgrade to 1.8.29.

StarlitLuna:
- Add DPI awareness at app start.
- Add setting to control whether DPI awareness is used.

### DIFF
--- a/src/LiveSplit.Core/Options/Settings.cs
+++ b/src/LiveSplit.Core/Options/Settings.cs
@@ -11,6 +11,7 @@ namespace LiveSplit.Options;
 
 public class Settings : ISettings
 {
+    public const int DPI_AWARENESS_OS_MIN_VERSION = 6;
     public IDictionary<string, HotkeyProfile> HotkeyProfiles { get; set; }
     public IList<RecentSplitsFile> RecentSplits { get; set; }
     public IList<string> RecentLayouts { get; set; }
@@ -24,6 +25,7 @@ public class Settings : ISettings
     public IList<RaceProviderSettings> RaceProvider { get; set; }
     public IList<string> ActiveAutoSplitters { get; set; }
     public IDictionary<string, bool> ComparisonGeneratorStates { get; set; }
+    public bool EnableDPIAwareness { get; set; }
 
     // Deprecated properties
     public KeyOrButton SplitKey
@@ -111,7 +113,8 @@ public class Settings : ISettings
             RefreshRate = RefreshRate,
             ServerPort = ServerPort,
             ActiveAutoSplitters = new List<string>(ActiveAutoSplitters),
-            ComparisonGeneratorStates = new Dictionary<string, bool>(ComparisonGeneratorStates)
+            ComparisonGeneratorStates = new Dictionary<string, bool>(ComparisonGeneratorStates),
+            EnableDPIAwareness = EnableDPIAwareness
         };
     }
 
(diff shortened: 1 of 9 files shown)


QueenSuzie:
Resolves #2563

QueenSuzie:
resolves #2563

QueenSuzie:
Considering no DPI aware has been the default since all time and is what most people have their streaming/recording layouts configured for, I made the setting false by default. I think this is reasonable since I *feel* like there aren't exactly a ton of people out there clamoring for this. But now the option exists and can be made aware of to people who need/want it.

src/LiveSplit.View/View/TimerForm.cs
```diff
@@ -2796,6 +2797,21 @@ private void LoadSettings()
         Settings = new StandardSettingsFactory().Create();
     }
 
+    private void SetDPIAwareness()
```
wooferzfg, line 2800: This should be called when the settings change, too?

QueenSuzie: No, you can only enable DPI awareness at the start of the application, and once enabled it can't be turned off. That's why there's a tooltip added for the new checkbox that LiveSplit needs to be restarted to change the setting.

src/LiveSplit.Core/Options/Settings.cs
```diff
@@ -11,6 +11,7 @@ namespace LiveSplit.Options;
 
 public class Settings : ISettings
 {
+    public const int DPI_AWARENESS_OS_MIN_VERSION = 6;
```
wooferzfg, line 14: Is this saying the minimum Windows version is Vista? I'm pretty sure LiveSplit doesn't run on XP anyway

src/LiveSplit.View/View/SettingsDialog.Designer.cs
```diff
@@ -695,42 +757,51 @@ private void InitializeComponent()
...
             // txtServerPort
             // 
             this.txtServerPort.Anchor = ((System.Windows.Forms.AnchorStyles)(((System.Windows.Forms.AnchorStyles.Top | System.Windows.Forms.AnchorStyles.Bottom) 
             | System.Windows.Forms.AnchorStyles.Right)));
-            this.txtServerPort.Location = new System.Drawing.Point(139, 6);
+            this.txtServerPort.Location = new System.Drawing.Point(186, 7);
+            this.txtServerPort.Margin = new System.Windows.Forms.Padding(4);
             this.txtServerPort.Name = "txtServerPort";
-            this.txtServerPort.Size = new System.Drawing.Size(51, 20);
+            this.txtServerPort.Size = new System.Drawing.Size(67, 22);
```
wooferzfg, line 773: There's a lot of unrelated size changes that shouldn't be here. I'm assuming it's because you didn't set the scaling in the Visual Studio Designer to 100%.

src/LiveSplit.Core/Options/Settings.cs
QueenSuzie: Yeah, DPI awareness isn't a thing until Vista. I think it doesn't hurt to have a check for this even if for no other reason than self documenting code.

src/LiveSplit.View/View/SettingsDialog.Designer.cs
QueenSuzie: Could be. Let me try to redo this part tomorrow.

QueenSuzie: Should be good now @wooferzfg

PR #2579, opened 2025-03-01
Always track gametime - remove GT if needed on reset

Oekn5w:
When the loading time edges are given via Named Pipe, TCP or Websocket server (via "pausegametime" and "unpausegametime"), GameTime would show "-" for any splits that don't have loads yet.

The gametime in these splits are `null`, so they cannot be used as expected for comparison.

This PR always saves the current GT, even if the loads are not initialized. If they are still not initialized when resetting, the GTs are removed, so the behavior for RT runs doesn't change, the GT tab will remain empty in the splits editor.

Behavior tested via Websocket pause and unpause

### DIFF
--- a/src/LiveSplit.Core/Model/LiveSplitState.cs
+++ b/src/LiveSplit.Core/Model/LiveSplitState.cs
@@ -123,8 +123,7 @@ public class LiveSplitState : ICloneable
             else
             {
                 curTime.GameTime = IsGameTimePaused
-                    ? GameTimePauseTime
-                    : curTime.RealTime - (IsGameTimeInitialized ? LoadingTimes : null);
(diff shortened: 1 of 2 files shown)


just-ero:
What problem does this solve? Your description is not clear to me. Please provide a real world example.

Furthermore, what other behavior does this affect?

Jujstme:
Not tracking the game time when it is uninitialized is likely an intended behavior.
Changing this should be done consciously, for example by setting `timer.IsGameTimeInitialized` to `true` in the `onStart` section of am autosplitter.
Keep in mind changing he default behavior in livesplit can be a breaking change for applications and scripts that use this to infer whether the game has a gametime timing method or not.

Oekn5w:
Yeah, adding that to the autosplitter is fair, closing this.

Issue #2359, opened 2023-08-25
Diablo IV - Live split does not recognize other categories than full game


M1PY:
Hello, I am M1PY and I am co-moderating Diablo IV on speedrun.com.

We created new sub-categories under the "levels" category for Diablo IV for Nightmare Dungeon Speedrunning.

However, these are not recognized by Livesplit. If we add a new category to "full game", Livesplit recognizes it immediately. Is there a way to make livesplit recognize the "levels" category and it's sub-categories?

Best regards
M1PY

LukeSaward1:
Are the categories under the individual levels section in SRC?
If so, LiveSplit does not support this.

wooferzfg:
Duplicate of https://github.com/LiveSplit/LiveSplit/issues/2075

Issue #2075, opened 2021-01-16
Feature Request: Individual Levels in Run Categories


slaurent22:
After selecting Game Category, the Run Category options are prefilled from the full-game level categories on speedrun.com. This enhancement would also add individual levels. For example, those from https://www.speedrun.com/hollowknight/individual_levels are not featured in the options. See screenshot:

![hk_livesplit](https://user-images.githubusercontent.com/74829867/104827535-c8775080-582c-11eb-8c14-db8237233111.PNG)

DeathHound6:
I think this would be a nice addition
It would integrate well with components like World Record, to automatically show correct information for the category being run

PR #2581, opened 2025-03-03
Add initial support for individual run levels

piratesephiroth:
attempting to solve #2075 and #2359

### DIFF
--- a/src/LiveSplit.Core/Model/Run.cs
+++ b/src/LiveSplit.Core/Model/Run.cs
@@ -17,6 +17,7 @@ public class Run : IRun, INotifyPropertyChanged
 {
     private string gameName;
     private string categoryName;
+    private string levelName;
 
     /// <summary>
     /// The name of the comparison used to save your Personal Best splits.
@@ -58,6 +59,20 @@ public class Run : IRun, INotifyPropertyChanged
             categoryName = value;
             Metadata.Refresh();
             TriggerPropertyChanged("CategoryName");
+        }
+    }
+
+    /// <summary>
+    /// Gets or sets the category of the run.
+    /// </summary>
+    public string LevelName
+    {
+        get => levelName;
+        set
+        {
+            levelName = value;
+            Metadata.Refresh();
+            TriggerPropertyChanged("LevelName");
         }
     }
 
--- a/src/LiveSplit.Core/Model/RunMetadata.cs
+++ b/src/LiveSplit.Core/Model/RunMetadata.cs
@@ -16,9 +16,12 @@ public class RunMetadata
 
     private string oldGameName;
     private string oldCategoryName;
+    private string oldLevelName;
     private Lazy<Game> game;
     private bool gameLoaded;
     private Lazy<Category> category;
+    private CategoryType categoryType;
+    private Lazy<Level> level;
     private Lazy<SpeedrunComSharp.Run> run;
     private string runId;
     private string platformName;
@@ -211,6 +214,15 @@ public class RunMetadata
         }
     }
 
+    public Level Level
+    {
+        get
+        {
+            Refresh();
+            return level.Value;
+        }
+    }
+
     public RunMetadata(IRun run)
     {
         LiveSplitRun = run;
@@ -218,6 +230,7 @@ public class RunMetadata
         CustomVariables = new Dictionary<string, CustomVariable>();
         game = new Lazy<Game>(() => null);
         category = new Lazy<Category>(() => null);
+        level = new Lazy<Level>(() => null);
         this.run = new Lazy<SpeedrunComSharp.Run>(() => null);
     }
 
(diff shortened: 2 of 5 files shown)


src/LiveSplit.Core/Model/Run.cs
```diff
@@ -61,6 +62,20 @@ public string CategoryName
         }
     }
 
+    /// <summary>
+    /// Gets or sets the category of the run.
+    /// </summary>
+    public string LevelName
+    {
```
wooferzfg, line 69: This would also need to be added to the splits file, which requires livesplit-core changes

src/LiveSplit.View/View/RunEditorDialog.cs
```diff
@@ -412,6 +451,46 @@ private void RefreshCategoryAutoCompleteList()
         });
     }
 
+    private void RefreshLevelAutoCompleteList()
+    {
+        Task.Factory.StartNew(() =>
+        {
+            try
+            {
+                string[] levelNames = ["Easy", "Medium", "Hard"];
```
wooferzfg, line 460: I don't think these placeholders are necessary

src/LiveSplit.Core/Model/Run.cs
piratesephiroth: ah, I knew I was forgetting something...

piratesephiroth: done now
LiveSplit/livesplit-core#860

CryZe:
Isn't an IL a kind of category? As in, why can you specify a category and a level name? Though I may be missing something completely.

just-ero:
I would also like to see how this integrates with speedrun.com's IL categories. What does this PR add in that regard?

piratesephiroth:
> Isn't an IL a kind of category? As in, why can you specify a category and a level name? Though I may be missing something completely.

Run Categories have 2 types:
- **Per-game**, The most common type of category, called "Full Game" by the site. They work by themselves.
- **Per-leve**l. These categories work together with a Level. A level is an arbitrarily defined parameter.  It can be a difficulty level, clearing the game using a single credit, picking a specific class or playable character, etc.

Take a look at Mortal Kombat Trilogy for example.
Here's the main "Full Game" categories:
image

Now the categorizes that show up when you select a Level:
image

I implemented the Level selection to work similarly to the site.
The "Run Level" selection determines which values you get in the "Run Category" listing.
A blank value will result in the Full Game categories.

CryZe:
They call this scope on speedrun.com apparently. So we probably should name it something similar.

piratesephiroth:
> They call this scope on speedrun.com apparently. So we probably should name it something similar.

Perhaps, but it seems that the scope just differentiates between Full game and Individual Levels.
The selection of a Level already determines the scope.

wooferzfg:
> Isn't an IL a kind of category? As in, why can you specify a category and a level name? Though I may be missing something completely.

Each level can have multiple categories. For example, if "Dragon Roost Cavern" is a level, the categories might be "Any%" and "100%".

src/LiveSplit.Core/Model/RunMetadata.cs
```diff
@@ -256,21 +269,21 @@ public void Refresh()
                     game = new Lazy<Game>(() => gameTask.Result);
 
                     Task platformTask = Task.Factory.StartNew(() =>
+                    {
+                        Game game = this.game.Value;
+                        if ((game != null && !game.Platforms.Any(x => x.Name == PlatformName)) || (gameLoaded && game == null))
                         {
-                            Game game = this.game.Value;
-                            if ((game != null && !game.Platforms.Any(x => x.Name == PlatformName)) || (gameLoaded && game == null))
```
wooferzfg, line 261: Please get rid of all the whitespace changes like this

piratesephiroth: You mean the extra indentation that those Task blocks had (the red boxes)?
I removed that because the Task block above them does not use it (the orange box)

image

just-ero:
Let me know if I'm missing anything, but can ILs not simply be added to the list of categories in the dropdown with the same name? Are IL sub-categories (like you're showing in the screenshot) not treated as variables?

wooferzfg:
> Let me know if I'm missing anything, but can ILs not simply be added to the list of categories in the dropdown with the same name? Are IL sub-categories (like you're showing in the screenshot) not treated as variables?

They're not variables on Speedrun.com:

Screenshot_20250305_082000_Chrome Beta

Screenshot_20250305_082053_Chrome Beta

just-ero:
Are sub-categories not variables?

src/LiveSplit.Core/Model/RunMetadata.cs
wooferzfg: If you'd like to fix this, it would be better to make a separate PR and then rebase this one once it's merged. Or at the very least, splitting the formatting changes in this PR into a separate commit

piratesephiroth: all right

src/LiveSplit.View/View/RunEditorDialog.cs
wooferzfg: Any% and 100% apply to a ton of games. The names of levels vary a lot more

piratesephiroth: OK, removed

PR #2583, opened 2025-03-08
Adding games to Splits Editor combobox: Make task incremental and cancelable

### LINKED ISSUE #2580: Splits Editor hangs for good 3-5 seconds when open
Splits Editor hangs for good 3-5 seconds when open. Most probably the reason is that it's loading all the game names every time. Can we have cursor by default placed somewhere and not load the game names unless the game name textbox is selected.

Oekn5w:
Adding all games at once seems to block the Split Editor Form, so it is unusable for around 5 seconds while the ~44k games are added to the combobox.

To fix the inability to edit the splits immediately after loading, the `AddRange` calls are split into segments of 1000 elements each. I didn't run a optimization on that value, but setting 10000 still visibly lags the Split Editor.

While this makes the splits editable directly after the Form opens, on pressing `Ok` we would still need to wait until all games have been added. Therefore the task is made cancelable and it is getting cancelled when the Form is getting closed.

Fixes #2580

### DIFF
--- a/src/LiveSplit.View/View/RunEditorDialog.cs
+++ b/src/LiveSplit.View/View/RunEditorDialog.cs
@@ -322,32 +322,7 @@ public partial class RunEditorDialog : Form
                 if (cachedGameNames != null)
                 {
                     gameNames = cachedGameNames.ToArray();
-                    this.InvokeIfRequired(() =>
-                    {
-                        bool remaining = true;
-                        int currentIdx = 0;
-                        try
-                        {
-                            while (remaining && !token.IsCancellationRequested)
-                            {
-                                Application.DoEvents();
-                                int count = Math.Min(gameNames.Count() - currentIdx, MAXADDITIONSPERADDRANGE);
-                                string[] subGameNames = new string[count];
-                                for (int i = 0; i < count; ++i)
-                                {
-                                    subGameNames[i] = gameNames[currentIdx + i];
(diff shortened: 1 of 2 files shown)


src/LiveSplit.View/View/RunEditorDialog.cs
```diff
@@ -320,10 +324,24 @@ private void FillCbxGameName()
...
                     this.InvokeIfRequired(() =>
                     {
-                        Application.DoEvents();
+                        bool remaining = true;
+                        int currentIdx = 0;
                         try
                         {
-                            cbxGameName.Items.AddRange(gameNames);
+                            while (remaining && !token.IsCancellationRequested)
+                            {
```
wooferzfg, line 332: Make this while loop a method that's also called below on `newGameNames`?

wooferzfg:
> Adding all games at once seems to block the Split Editor Form, so it is unusable for around 5 seconds while the ~44k games are added to the combobox.

This isn't the case for all users, by the way (like it doesn't happen for me). Not sure why it varies

src/LiveSplit.View/View/RunEditorDialog.cs
Oekn5w: I wasn't sure that would work because I never dabbled into C# Threading, but it looks like it works

PR #2587, opened 2025-03-17
Adds HCPComparisonGenerator

rodrigosiqueiradev:
HCP is a way of calculating results in Golf. It's based of the last 20 rounds played and averaging the top 8 results in that set.

This is an application of this method to give runners a goal different from beating their PBs, which tends to be the case for a lot of runners. This new HCP goal aims for the runner to be able to improve their overall timings, or handicap as it's commonly referred to, without the common mental pressure of beating their best time ever, which can demotivate a lot of folks when they don't see their skills evolving.

### DIFF
--- a/src/LiveSplit.Core/Model/Comparisons/HCPComparisonGenerator.cs
+++ b/src/LiveSplit.Core/Model/Comparisons/HCPComparisonGenerator.cs
@@ -1,0 +1,98 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+using LiveSplit.Options;
+
+using static System.Math;
+
+namespace LiveSplit.Model.Comparisons;
+
+public class HCPComparisonGenerator : IComparisonGenerator
+{
+    public IRun Run { get; set; }
+    public const string ComparisonName = "Golf HCP";
+    public const string ShortComparisonName = "HCP";
+    public string Name => ComparisonName;
+    private const int NumRuns = 8;
+    private const int AverageSamplesCount = 20;
+
+    public HCPComparisonGenerator(IRun run)
+    {
+        Run = run;
+    }
+
+    protected TimeSpan CalculateAverage(IEnumerable<TimeSpan> curList)
+    {
+        double averageTime = curList.OrderBy(x => x.TotalSeconds).Take(AverageSamplesCount).ToList().Aggregate(0.0, (s, x) => s + x.TotalSeconds) / Min(AverageSamplesCount, curList.Count());
+        return TimeSpan.FromTicks((long)(averageTime * TimeSpan.TicksPerSecond));
+    }
+
+    public void Generate(TimingMethod method)
+    {
+        var allHistory = new List<List<TimeSpan>>();
+        foreach (ISegment segment in Run)
+        {
+            allHistory.Add([]);
+        }
+
+        // Get the last 20 complete runs 
+        var recentRuns = Run.AttemptHistory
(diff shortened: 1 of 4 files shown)


src/LiveSplit.Core/Model/Comparisons/HCPComparisonGenerator.cs
```diff
@@ -0,0 +1,98 @@
...
+
+namespace LiveSplit.Model.Comparisons;
+
+public class HCPComparisonGenerator : IComparisonGenerator
+{
+    public IRun Run { get; set; }
+    public const string ComparisonName = "Golf HCP";
+    public const string ShortComparisonName = "HCP";
+    public string Name => ComparisonName;
+    private const int NumRuns = 8;
```
wooferzfg, line 17: Rename to `MaximumNumberOfBestRunsToIncludePerSegment`?

It's worth noting that you're taking the latest 20 runs, but the top 8 runs is being applied per segment, rather than using the 8 fastest runs overall. Is that intended?

src/LiveSplit.Core/Model/Comparisons/HCPComparisonGenerator.cs
```diff
@@ -0,0 +1,98 @@
...
+    private const int AverageSamplesCount = 20;
```
wooferzfg, line 18: Rename to `NumberOfLatestRunsToInclude`?

src/LiveSplit.Core/Model/Comparisons/HCPComparisonGenerator.cs
rodrigosiqueiradev: Yes, the decision of only considering the top 8 runs over the last 20 was per design as even within the same segment, sometimes strategies change. I'll apply the name change as soon as possible.

src/LiveSplit.Core/Model/Comparisons/HCPComparisonGenerator.cs
rodrigosiqueiradev: Will do

rodrigosiqueiradev:
Apologies if re-requesting review was not needed, wasn't sure of the process so I went the extra click just for the sake of it

src/LiveSplit.Core/Model/Comparisons/HCPComparisonGenerator.cs
wooferzfg: To double check, it's intended that top 8 is per segment, but latest 20 is *not* per segment?

src/LiveSplit.Core/Model/Comparisons/HCPComparisonGenerator.cs
```diff
@@ -0,0 +1,98 @@
...
+    private const int NumberOfLatestRunsToInclude = 20;
+
+    public HCPComparisonGenerator(IRun run)
+    {
+        Run = run;
+    }
+
+    protected TimeSpan CalculateAverage(IEnumerable<TimeSpan> curList)
+    {
+        double averageTime = curList.OrderBy(x => x.TotalSeconds).Take(MaximumNumberOfBestRunsToIncludePerSegment).ToList().Aggregate(0.0, (s, x) => s + x.TotalSeconds) / Min(MaximumNumberOfBestRunsToIncludePerSegment, curList.Count());
```
wooferzfg, line 27: Use this? https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.average?view=net-9.0

rodrigosiqueiradev: My lack of knowledge from C# is showing... Using that function now

Issue #2588, opened 2025-03-30
[Feature Request] Add ability to pull various Completed and Attempt Run Count information from the server


SushiKishi:
I wanted to be able to pull the total number of attempts from the LiveSplit server in order to add a custom "label" to it.  Since my stream has a fighting / wrestling style layout, each run is a different "Round" like in a boxing match, and there was no way to simply get LiveSplit to display "Round {TotalAttempts}" in the layout.

And I figured out enough to be able to add that and build it on my own, so I added a few other commands as well.

I'm also completely pants at doing anything in GitHub; I can't for the live of me figure out how to submit a pull request adding features, but apparently I can do so if I first create an issue, then hopefully can attach my changes to this issue as a pull request.

So the request would be to add the following outputs/commands to the LiveSplit server component:

1. Total Attempts
2. Completed Runs
3. {CompletedRuns}/{TotalAttempts}
4. Win Rate, formatted as a percentage (i.e. 3/4 = 75.0)
5. Win Rate, formatted as a decimal (i.e. 3/4 = 0.750)

wooferzfg:
#1 and #2 make sense to add; #3/#4/#5 can all be derived from the first two so you don't need those

PR #2589, opened 2025-03-30
Add Completed and Attempted Run Count information to LiveSplit Server command list

SushiKishi:
For Issue #2588.

Adds the following commands available for the Server component:

- getattemptcount
    -  Gets the total number of attempted runs, including the current run (as it is updated at start)
- getcompletedcount
    -  Gets the total number of COMPLETED runs, using the same method that the "Title" layout component does (`State.Run.AttemptHistory.Count(x => x.Time.RealTime != null)`)

I've been able to test this using PHP to access the TCP server. But I'm not smart enough to program something to access -anything- in the LiveSplit server via sockets or pipes, so I'd appreciate some help in making sure it works as intended that way, as well.

The following was removed from the pull request as originally submitted:
- A command to output `{CompletedRuns}/{AttemtpedRuns}`
- Commands to output the winrate (e.g. 3 completed runs, 4 attempted runs, output either `75.0` or `.750`

### DIFF
--- a/src/LiveSplit.Core/Server/CommandServer.cs
+++ b/src/LiveSplit.Core/Server/CommandServer.cs
@@ -506,6 +506,11 @@ public class CommandServer
                 response = "pong";
                 break;
             }
+            case "getattemptcount":
+            {
+                response = Model.CurrentState.Run.AttemptCount.ToString();
+                break;
+            }
             default:
             {
                 Log.Error($"[Server] Invalid command: {message}");
--- a/src/LiveSplit.Core/Server/CommandServer.cs
+++ b/src/LiveSplit.Core/Server/CommandServer.cs
@@ -544,7 +544,7 @@ public class CommandServer
 
                 else
                 {
-                    response = Math.Round(State.Run.AttemptHistory.Count(x => x.Time.RealTime != null) / (double)Model.CurrentState.Run.AttemptCount, 3).ToString();
+                    response = Math.Round(State.Run.AttemptHistory.Count(x => x.Time.RealTime != null) / (double)Model.CurrentState.Run.AttemptCount, 3).ToString().TrimStart('0');
                 }
                 break;
             }
--- a/src/LiveSplit.Core/Server/CommandServer.cs
+++ b/src/LiveSplit.Core/Server/CommandServer.cs
@@ -523,12 +523,29 @@ public class CommandServer
             }
             case "getwinratepct":
             {
-                response = Math.Round((State.Run.AttemptHistory.Count(x => x.Time.RealTime != null) / (double)Model.CurrentState.Run.AttemptCount) * 100, 1).ToString();
+                if (Model.CurrentState.Run.AttemptCount == 0)
+                {
+                    response = "0.0";
+                }
+
+                else
+                {
+                    response = Math.Round((State.Run.AttemptHistory.Count(x => x.Time.RealTime != null) / (double)Model.CurrentState.Run.AttemptCount) * 100, 1).ToString();
+                }
+                
                 break;
             }
             case "getwinratedec":
             {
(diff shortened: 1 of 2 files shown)


src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -506,6 +506,48 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+            case "getwinratepct":
+            {
+                if (Model.CurrentState.Run.AttemptCount == 0)
+                {
+                    response = "0.0";
+                }
+
+                else
+                {
+                    response = Math.Round((State.Run.AttemptHistory.Count(x => x.Time.RealTime != null) / (double)Model.CurrentState.Run.AttemptCount) * 100, 1).ToString();
```
wooferzfg, line 533: Why can't you just do `(completed count) / (total count)`

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -506,6 +506,48 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                break;
+            }
+			case "getcompletedcount":
+            {
+                response = State.Run.AttemptHistory.Count(x => x.Time.RealTime != null).ToString();
+                break;
+            }
+            case "getcompletionsoverattempts":
+            {
+                response = State.Run.AttemptHistory.Count(x => x.Time.RealTime != null).ToString() + "/" + Model.CurrentState.Run.AttemptCount.ToString();
```
wooferzfg, line 521: This is unnecessary because it can be derived from the two commands above

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -506,6 +506,48 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                }
+
+                else
+                {
+                    response = Math.Round((State.Run.AttemptHistory.Count(x => x.Time.RealTime != null) / (double)Model.CurrentState.Run.AttemptCount) * 100, 1).ToString();
+                }
+                
+                break;
+            }
+            case "getwinratedec":
```
wooferzfg, line 538: This is unnecessary

wooferzfg (changes requested):

src/LiveSplit.Core/Server/CommandServer.cs
SushiKishi: Should it just be `getattemptcount` and `getcompletedcount` then?

I had left off the percent sign for the win percentage assuming whoever was needing it could add it on their own, but that could be applied to the other options too. If you can just get the completions and attempts you'd be able to generate the other options.

src/LiveSplit.Core/Server/CommandServer.cs
SushiKishi: I might be misunderstanding the question (and it might be moot if I'm just adding Completions and Attempts as commands), but:

Completed Count / Total Count would generate the decimal (i.e. 3 complete runs, 4 attempts, output is 0.750), where this was meant to output the percentage (75.0)

wooferzfg: The program that's using the server can handle formatting concerns like this

src/LiveSplit.Core/Server/CommandServer.cs
wooferzfg: > Should it just be `getattemptcount` and `getcompletedcount` then?

Yes, the other data can be derived from those

wooferzfg:
Please also update the commands list in the README

SushiKishi:
I updated the command list!

GitHub's also telling me to report on a build check, is there something I need to do on my end with that?

CryZe:
No, we just need to approve it manually.

PR #2596, opened 2025-04-09
New Run Info Server Commands

### LINKED ISSUE #2594: [Feature Request] LiveSplit Server commands to fetch some run info
Planning for a project in the future involving the LiveSplit Server and I will need commands to gather game + category names as they will be used as checks to help prevent unexpected program behavior. A command to return a string of the current comparison would just be a nice to have as its unrelated for this. As I found the code that could be used to easily fetch everything I need, I might be able to work on this feature myself once I can sit down and setup my environment to fork LiveSplit.

Names of commands + code:
1. 'getcurrentgamename' Returns the name of current game - `CurrentState.Run.GameName;`
2. 'getcurrentcategoryname' Returns name of current category - `CurrentState.Run.CategoryName;`
3. 'getcurrentcomparisonname' Returns the name of current comparison from current run - `State.CurrentComparison;`
4. 'getcurrentextendedcategoryname' Returns extended name of current category that includes additional category info - `state.Run.GetExtendedCategoryName(Settings.ShowRegion, Settings.ShowPlatform, Settings.ShowVariables);`

Note: for category names, if only the category name (excluding variables, region, platform, etc.) can easily be extracted from 'getcurrentextendedcategoryname' no matter the game + category, this will eliminate the need for the 'getcurrentcategoryname' command or it could the shortened command name to fetch everything.

aop-runs:
CLOSES #2594

Adds the following server commands:

1. getcomparisonname - Current comparison name
2. getgamename - Current game name
3. getcategoryname - Current category name
4. gettimingmethod - Current timing method
5. getsplitcount - Current split count
6. getcategoryvariables - Current region + platform + emulator usage + variables
7. getupcomingsplitname - Next split name from current split (Returns '-' on final split)
8. getpausedrealtime - Current time paused using real timing method
9. getpausedgametime - Current time paused using game timing method
10. getoffset - Offset for the starting time of current run
11. gethotkeyprofile - Current name of hotkey profile
12. getlivesplitversion - Current LiveSplit version installed
13. getlivesplitpath - Gathers absolute filepath of LiveSplit executable


src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +291,20 @@ private void ProcessMessage(string message, IConnection clientConnection)
                 State.IsGameTimePaused = true;
                 break;
             }
+            case "getgamename":
+            case "getcurrentgamename":
```
wooferzfg, line 295: Just `getgamename` is fine. Same with `getcategoryname` and `getcurrentcomparisonname`

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +291,20 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+            {
+                response = State.Run.GameName.ToString();
+                break;
+            }
+            case "getcategoryname":
+            case "getcurrentcategoryname":
+            {
+                String category = State.Run.CategoryName.ToString();
+                String extendedcategory = State.Run.GetExtendedCategoryName(true, true, true).ToString();
```
wooferzfg, line 304:
```suggestion
                var extendedCategoryName = State.Run.GetExtendedCategoryName(true, true, true).ToString();
```

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +291,20 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                response = (category.Length != extendedcategory.Length) ? extendedcategory.Insert(category.Length, " -") : category;
```
wooferzfg, line 305: Why not just always return the extended category name? Otherwise you're repeating the category twice

wooferzfg: If your use case needs `Any%`, then make the category name a separate command from the extended category name

src/LiveSplit.Core/Server/CommandServer.cs
wooferzfg: This is to match what they're called in the LiveSplit UI - "Game Name", "Category Name", and "Current Comparison"

aop-runs: It was due to other commands having alternate names, but I can change it easily.

src/LiveSplit.Core/Server/CommandServer.cs
aop-runs: Wasn't thinking about that tbh so thanks for the heads up. I wanted to do it all in one command to avoid redundancy, but it seems like 2 might be the only way after all.

aop-runs:
> CLOSES #2594
> 
> Adds the following server commands:
> 
> ~1. getcurrentcomparisonname - Current comparison name 2. getcurrentgamename - Current game name 3. getcurrentcategoryname - Current category name with variables (if any exist)~
> 
> ~Note: If variables or other details are present in category name, they are separated with ' - ' for easy extraction.~
> 
> 1. getcomparisonname - Current comparison name
> 2. getgamename - Current game name
> 3. getcategoryname - Current category name
> 4. gettimingmethod - Current timing method
> 5. getsplitcount - Current split count
> 6. getcategoryvariables - Current variables + region + platform

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +291,96 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                    String variablePairs = $"{{{string.Join(", ", vKeys)}}}";
+                    keyL.Add($"\"variables\": {variablePairs}");
+                }
+                //Combine
+                String keys = "";
+                if (keyL.Any())
+                {
+                    keys = string.Join(", ", keyL);
+                }
+                response = $"{{{keys}}}";
```
just-ero, line 381:
```suggestion
                response = "{" + string.Join(", ", keyL) + "}";
```
This should really be done with a `StringBuilder`.

just-ero:
Your editor should be throwing a good amount of style warnings at you. Please don't ignore them.

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +291,96 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+            case "getcategoryname":
+            {
+                response = State.Run.CategoryName.ToString();
+                break;
+            }
+            case "getcategoryvariables":
+            {
+                var keyL = new List<string>();
+                //Region
+                bool doSimpleRegion = !State.Run.Metadata.GameAvailable;
```
just-ero, line 308: What is this check for? Does `Metadata.Region` not exist just because the game doesn't exist?

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +291,96 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                if (doSimpleRegion)
+                {
+                    if (!string.IsNullOrEmpty(State.Run.Metadata.RegionName))
+                    {
+                        keyL.Add($"\"region\": \"{State.Run.Metadata.RegionName}\"");
+                    }
+                }
+                else if (State.Run.Metadata.Region != null && !string.IsNullOrEmpty(State.Run.Metadata.Region.Abbreviation) && State.Run.Metadata.Game != null && State.Run.Metadata.Game.Regions.Count > 1)
```
just-ero, line 316: Why does `State.Run.Metadata.Game.Regions.Count` need to be `> 1` when we return `State.Run.Metadata.Region.Abbreviation`?

just-ero, line 316:
```suggestion
                else if (md is { Region.Abbreviation.Length: > 0, Game.Regions.Count: > 1 })
```

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +291,96 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                {
+                    keyL.Add($"\"region\": \"{State.Run.Metadata.Region.Abbreviation}\"");
+                }
+                //Platform
+                bool doSimplePlatform = !State.Run.Metadata.GameAvailable;
```
just-ero, line 321: What is this check for? Does `Metadata.PlatformName` not exist just because the game doesn't exist?

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +291,96 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                if (!string.IsNullOrEmpty(State.Run.Metadata.PlatformName) && (doSimplePlatform || (State.Run.Metadata.Game != null && State.Run.Metadata.Game.Platforms.Count > 1)))
```
just-ero, line 322:
```suggestion
                if (md is { PlatformName.Length: > 0, Game: null or { Platforms.Count: > 1 } })
```

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +291,96 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                IEnumerable<string> variables = State.Run.Metadata.VariableValueNames.Keys;
+                if (State.Run.Metadata.GameAvailable && State.Run.Metadata.Game != null)
+                {
+                    string categoryId = null;
+                    if (State.Run.Metadata.CategoryAvailable && State.Run.Metadata.Category != null)
+                    {
+                        categoryId = State.Run.Metadata.Category.ID;
+                    }
+                    variables = State.Run.Metadata.Game.FullGameVariables.Where(x => x.CategoryID == null || x.CategoryID == categoryId).Select(x => x.Name);
+                }
```
just-ero, line 348:
```suggestion
                IEnumerable<string> variables = State.Run.Metadata.Game?.FullGameVariables
                    .Where(v => v.CategoryID == State.Run.Metadata.Category?.ID)
                    .Select(v => v.Name)
                    ?? State.Run.Metadata.VariableValueNames.Keys;
```

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +291,96 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                        else if (valueLower == "no")
+                        {
+                            vKeys.Add($"\"{variable}\": \"No {name}\"");
+                        }
+                        else
+                        {
+                            vKeys.Add($"\"{variable}\": \"{variableValue}\"");
+                        }
+                    }
+                }
```
just-ero, line 369: If we're already formatting this like JSON, we should probably just introduce a format like
```json
{
  "name": "<variable-name>",
  "value": "<variable-value>"
}
```

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +291,96 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                    {
+                        categoryId = State.Run.Metadata.Category.ID;
+                    }
+                    variables = State.Run.Metadata.Game.FullGameVariables.Where(x => x.CategoryID == null || x.CategoryID == categoryId).Select(x => x.Name);
+                }
+                foreach (string variable in variables)
+                {
+                    if (State.Run.Metadata.VariableValueNames.ContainsKey(variable))
+                    {
+                        string name = variable.TrimEnd('?');
```
just-ero, line 353: Why is this being trimmed?

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +291,96 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                        string variableValue = State.Run.Metadata.VariableValueNames[variable];
+                        string valueLower = variableValue.ToLowerInvariant();
+                        if (valueLower == "yes")
+                        {
+                            vKeys.Add($"\"{variable}\": \"{name}\"");
+                        }
+                        else if (valueLower == "no")
+                        {
+                            vKeys.Add($"\"{variable}\": \"No {name}\"");
+                        }
```
just-ero, line 363:
```suggestion
                        if (variableValue.Equals("yes", StringComparison.OrdinalIgnoreCase))
                        {
                            vKeys.Add($"\"{variable}\": \"{name}\"");
                        }
                        else if (variableValue.Equals("no", StringComparison.OrdinalIgnoreCase))
                        {
                            vKeys.Add($"\"{variable}\": \"No {name}\"");
                        }
```

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +291,96 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                        else
+                        {
+                            vKeys.Add($"\"{variable}\": \"{variableValue}\"");
+                        }
```
just-ero, line 367: I don't agree with special handling for variables with question marks and Yes/No options.

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +291,96 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                    }
+                }
+                if (vKeys.Any())
```
just-ero, line 370: Avoid multiple enumeration.

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +291,96 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                {
+                    String variablePairs = $"{{{string.Join(", ", vKeys)}}}";
+                    keyL.Add($"\"variables\": {variablePairs}");
+                }
+                //Combine
+                String keys = "";
+                if (keyL.Any())
```
just-ero, line 377: Avoid multiple enumeration.

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -314,6 +404,12 @@ private void ProcessMessage(string message, IConnection clientConnection)
                 response = splitindex.ToString();
                 break;
             }
+            case "getsplitcount":
+            {
+                int splitcount = State.Run.Count;
+                response = splitcount.ToString();
```
just-ero, line 410:
```suggestion
                response = State.Run.Count.ToString();
```

aop-runs:
@just-ero The approach for gathering everything was based off of (https://github.com/LiveSplit/LiveSplit/blob/0e9d46c029246b82edbda9c58739f3058e03a3f5/src/LiveSplit.Core/Model/IRun.cs#L391) mainly used for the title component which gave me everything I needed for this command including checks for what data exists. Also chose to make a JSON string for this specifically to leave out unavailable data and let others choose only what they need. I do agree with your concerns though.

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -303,39 +304,39 @@ private void ProcessMessage(string message, IConnection clientConnection)
             }
             case "getcategoryvariables":
             {
-                var keyL = new List<string>();
+                Dictionary<string, Object> keys = new Dictionary<string, Object>();
```
just-ero, line 307: Create a concrete type to serialize instead. Never use `object`, and especially not `Object` : https://blog.paranoidcoding.org/2019/04/08/string-vs-String-is-not-about-style.html.

```cs
file sealed record CategoryMetadata(
  [property: JsonPropertyName("region")] string Region,
  string Platform,
  CategoryVariable[] Variables);

file sealed record CategoryVariable(
  string Name,
  string Value);
```
```cs
JsonSerializer.Serialize(metadata, new() { PropertyNamingPolicy = JsonNamingPolicy.CamelCase });
```

aop-runs: I'll admit that I'm not that fluent in C# compared to other languages so feedback is always appreciated. I'm learning it specifically to contribute for the LS server since I need some commands for a future project anyways.

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -30,6 +32,7 @@ public class CommandServer
     protected NamedPipeServerStream WaitingServerPipe { get; set; }
 
     protected bool AlwaysPauseGameTime { get; set; }
+    protected record CategoryMetaData(string Region, string Platform, IReadOnlyDictionary<string, string> Variables);
```
just-ero, line 35: Since this type is used only locally for the serialization, either mark it `private` and keep it here, or mark it `file` and place it outside of the class (preferrably at the end of the file).

src/LiveSplit.Core/Server/CommandServer.cs
just-ero: @wooferzfg

src/LiveSplit.Core/Server/CommandServer.cs
just-ero: @wooferzfg

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +294,94 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                        }
+                        else if (valueLower == "no")
+                        {
+                            varDict.Add(variable, $"No {name}");
+                        }
+                        else
+                        {
+                            varDict.Add(variable, variableValue);
+                        }
+                    }
```
just-ero, line 374:
```suggestion
                    if (State.Run.Metadata.VariableValueNames.TryGetValue(variable, out string value))
                    {
                        varDict.Add(variable, value);
                    }
```

Is the special handling for `Yes` and `No` really necessary?

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +294,94 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                }
+                //Combine
+                CategoryMetaData metadata = new CategoryMetaData(region, platform, varDict.Count() > 0 ? varDict : null);
```
just-ero, line 377:
```suggestion
                CategoryMetaData metadata = new(region, platform, varDict.Count() > 0 ? varDict : null);
```

Your editor should have suggested this change to you. Please don't ignore style-related diagnostics.

src/LiveSplit.Core/Server/CommandServer.cs
just-ero, line 35:
```suggestion
    protected record CategoryMetadata(string Region, string Platform, IReadOnlyDictionary<string, string> Variables);
```

src/LiveSplit.Core/Server/CommandServer.cs
aop-runs: Without them, the values for the variable would be displayed as yes or no just like the title component. I don't think there's a way around this scenario all together, but at least the key will have the variable name for the command so I'll do away with it.

src/LiveSplit.Core/Server/CommandServer.cs
aop-runs: I did not get something for this and I'm on the VS setup from the README.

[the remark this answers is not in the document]
aop-runs: This was to leave out the variables from the response all together if it the dictionary was empty. If I include it despite having no values, I would include all metadata for consistency even if nothing is present.

just-ero: On a completely fresh install of Visual Studio and a fresh clone of your branch, I do get these diagnostics:
image
image

aop-runs: I've should clarify that particular change wasn't suggested, but I did get some that were likely worse. Sorry about that!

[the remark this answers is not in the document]
just-ero: I don't quite follow. Is there an issue in returning `{}` over `null`? `null` to me conveys that something went wrong. `{}` conveys that there simply are no variables.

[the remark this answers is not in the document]
aop-runs: Before the recent commit a few minutes go, I intended to set the dictionary value to null if it was empty since returning an empty dictionary seemed unneeded.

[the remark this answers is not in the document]
just-ero: The check for emptiness is unnecessary. Not returning something empty. Let's be as clear as possible with the data we return. If there are 0 things of something, it means the collection is empty, not `null`.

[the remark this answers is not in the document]
aop-runs: Fair enough, I did initially feel that it was better to keep empty/null data out of the response, but I already changed it to return everything regardless.

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -29,7 +30,11 @@ public class CommandServer
     protected ITimeFormatter TimeFormatter { get; set; }
     protected NamedPipeServerStream WaitingServerPipe { get; set; }
 
-    protected bool AlwaysPauseGameTime { get; set; }
+    private bool AlwaysPauseGameTime { get; set; }
```
just-ero, line 33: Why is this changed to `private`?

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +296,58 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                    platform = md.PlatformName;
+                }
+
+                //Variables
+                Dictionary<string, string> variables = [];
+                IEnumerable<string> variableL = md.VariableValueNames.Keys;
+                if (md.GameAvailable && md.Game != null)
+                {
+                    string categoryId = (md.CategoryAvailable && md.Category != null) ? md.Category.ID : null;
+                    variableL = md.Game.FullGameVariables.Where(x => x.CategoryID == null || x.CategoryID == categoryId).Select(x => x.Name);
```
just-ero, line 337:
```suggestion
                    variableL = md.Game.FullGameVariables.Where(fgv => fgv.CategoryID == md.Category?.ID).Select(fgv => fgv.Name);
```

* `md.CategoryAvailable` and `md.Category != null` check the same thing. It's not possible for only one to be true.
* Use more fitting lambda parameter names. `x` is not acceptable.
* Do we really want to include variables where `CategoryID` is null?
* `fgv.CategoryID == md.Category?.ID` checks the full game variable's `CategoryID` against either `null` if `md.Category` was `null`, or `md.Category.ID` if it wasn't. It's simply more concise.

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -291,6 +296,58 @@ private void ProcessMessage(string message, IConnection clientConnection)
...
+                response = State.Run.CategoryName.ToString();
+                break;
+            }
+            case "getcategoryvariables":
+            {
+                RunMetadata md = State.Run.Metadata;
+
+                //Region
+                string region = null;
+                if (md.Region != null && !string.IsNullOrEmpty(md.Region.Abbreviation) && md.Game != null)
```
just-ero, line 315: Any particular reason as to why you don't want to accept my pattern matching suggestions here?
```suggestion
                if (md is { Game: not null, Region.Length: > 0 })
```

just-ero:
Other than that, the PR looks complete to me.

src/LiveSplit.Core/Server/CommandServer.cs
aop-runs: Was only used in this class. Didn't cause new issues testing the relevant commands, but I may change it back for consistency.

just-ero: You must change this back. This would be a binary breaking change.

src/LiveSplit.Core/Server/CommandServer.cs
```diff
@@ -7,6 +7,7 @@
 using System.Net.Sockets;
 using System.Text.Json;
 using System.Text.RegularExpressions;
+using System.Threading;
```
just-ero: For what purpose?

wooferzfg:
Why do you have changes to submodules?

aop-runs:
> Why do you have changes to submodules?

These were changes from other commits during May-June that got merged to master since I initially created this PR. I just updated my fork with these changes to avoid merge conflicts.

PR #2599, opened 2025-04-24
CommandServer: add setcustomvariable

### LINKED ISSUE #2574: Allow setting CustomVariable via server
Add a command `setcustomvariablevalue NAME VALUE` that lets a client create or update a `CustomVariable`. This will work hand in hand with #2565 and open up so many possibilities.

#### Personal Use Case
I develop and maintain AutoSplit64 (autosplitter for Super Mario 64 on console), which controls LiveSplit via the server. AutoSplit64 maintains its own route files to control splitting behaviour, and I use a custom variable to associate an AS64 route ID with one or more LSS files. Then, using `getcustomvariablevalue`, I automatically keep AS64 sync'd with LiveSplit. I use a work-around at the moment to set the variable, but this command would really make the LiveSplit server feel complete, and allow server-using programs to integrate far more seamlessly.

AlexKnauth:
Resolves #2574 by adding `setcustomvariable` to let a client create or update a `CustomVariable`.

Expects an argument encoding using JSON to combine 2 strings into 1 string.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ Commands that generate no response:
 - switchto gametime
 - setsplitname INDEX NAME
 - setcurrentsplitname NAME
+- setcustomvariable NAME VALUE
 
 Commands that return a time:
 
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ Commands that generate no response:
 - switchto gametime
 - setsplitname INDEX NAME
 - setcurrentsplitname NAME
-- setcustomvariable NAME VALUE
+- setcustomvariable JSON([NAME, VALUE])
 
 Commands that return a time:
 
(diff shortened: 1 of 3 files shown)


AlexKnauth:
Perhaps encoding the 2 strings into 1 string using JSON would be better.

AlexKnauth:
I have changed it to expect and argument encoding using JSON to combine 2 strings into 1 string, and tested that using this toy client:

```python
import json
import socket

s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
print(s.connect(("localhost", 16834)))
print(s.send(b"starttimer\n"))
print(s.send(b"setcustomvariable [\"segment hits\", \"fate\"]\n"))
print(s.send(b"setcustomvariable [\"hits\", \"as punishment\"]\n"))

def setcustomvariable(name, value):
    print(s.send(b"setcustomvariable " + json.dumps([name, value]).encode() + b"\n"))

setcustomvariable("delta hits", "worse")
setcustomvariable("pb hits", "expectations")
```

Using the `custom_variable_splits.lsl` layout from here: https://github.com/LiveSplit/livesplit-core/pull/874/files#diff-7c44c9d733946bd951ac93b676fda34928d17d32e9cda2ba038d90e6129159b6

PR #2600, opened 2025-04-26
Segment: CustomVariableValues

AlexKnauth:
A preparation PR for https://github.com/LiveSplit/LiveSplit.Splits/pull/27 to use to display Custom Variables in a column of the splits component. This will allow Custom Variable values to persist across layout changes.

### DIFF
--- a/src/LiveSplit.Core/Model/ISegment.cs
+++ b/src/LiveSplit.Core/Model/ISegment.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Drawing;
 
 using LiveSplit.Model.Comparisons;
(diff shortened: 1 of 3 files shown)


wooferzfg:
These variables cannot be permanent?

AlexKnauth:
When a Custom Variable is set to permanent, its value is currently stored in only one place: the `CustomVariables` section of the metadata in the splits file, like this:
```
    <Variables>
      <Variable name="Patch">1.5.78</Variable>
    </Variables>
    <CustomVariables>
      <Variable name="pb hits">14</Variable>
    </CustomVariables>
```

If the custom variable has different values at different split points during the run, those are not stored: this is consistent with how LiveSplit One handles them.

CryZe:
To clarify: It's not intentional that it's not persisted further. We just need to actually implement it at some point.

PR #2610, opened 2025-06-27
Fixes calculation logic on HCPComparisonGenerator 

rodrigosiqueiradev:
Corrects implementation on comparison generator to calculate HCP on the final time of the run instead off calculating each individual segment. Segment times are filled in with scaled values that sums up to the calculated final time.

### DIFF
--- a/src/LiveSplit.Core/Model/Comparisons/HCPComparisonGenerator.cs
+++ b/src/LiveSplit.Core/Model/Comparisons/HCPComparisonGenerator.cs
@@ -28,63 +28,97 @@ public class HCPComparisonGenerator : IComparisonGenerator
 
     public void Generate(TimingMethod method)
     {
-        var allHistory = new List<List<TimeSpan>>();
-        foreach (ISegment segment in Run)
+        int segmentCount = Run.Count;
+        var recentRuns = Run.AttemptHistory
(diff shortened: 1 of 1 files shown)


src/LiveSplit.Core/Model/Comparisons/HCPComparisonGenerator.cs
```diff
@@ -28,63 +28,97 @@ protected TimeSpan CalculateAverage(IEnumerable<TimeSpan> curList)
...
-                totalTime += CalculateAverage(curList);
+                double segTime = segmentTimesPerRun[run][seg];
+                double total = totalTimes[run];
+                if (total > 0)
+                {
+                    sum += segTime / total;
+                    count++;
+                }
             }
+            avgProportions[seg] = count > 0 ? sum / count : 1.0 / segmentCount;
```
wooferzfg, line 106: What if count is 0 here?

rodrigosiqueiradev: It would mean that there were no valid runs for this segment (perhaps if runs had missing or zero total times). This fallback evenly distributes the total time across all segments that were missing. Thought it would be safer to include some fallback value. You think I should've taken some other approach for that?

wooferzfg: Sorry, I misread and missed that you're explicitly handling the count = 0 case on this line

PR #2642, opened 2025-11-03
Use corrected refresh rate calulcation

bcoopers:
Current refresh rate calculation assumes the main loop takes 0 milliseconds. Correctly account for the time the main loop takes when deciding how long to sleep.


CryZe:
The way I usually like to to this is to simply added the refresh rate's time span to the previous time stamp and sleep until then. That way you only need to take one time stamp per iteration and technically don't lose any precision even between the time stamps. Also using an actual calendar based date time instead of a StopWatch is probably dangerous.

wooferzfg:
This looks similar to an old PR https://github.com/LiveSplit/LiveSplit/pull/2236

bcoopers:
> This looks similar to an old PR #2236

It looks like the issue with that PR was that there was still variance in refresh rate calculation due to the Windows scheduler, and the fix for it was extremely resource intensive.

While it doesn't resolve that issue, it still seems like a monotonic improvement over the current implementation, which ignores the main loop body runtime in its refresh calculation

AITYunivers:
You shouldn't be initializing a brand new stopwatch for each frame. Just make one and just Restart it

bcoopers:
> You shouldn't be initializing a brand new stopwatch for each frame. Just make one and just Restart it

sure. done

bcoopers:
Hello, anything else you'd like to see for this PR? WDYT?

wooferzfg:
> Hello, anything else you'd like to see for this PR? WDYT?

What impact does this have on the refresh rate in practice with the default settings? Does this affect autosplitters at all?

PR #2652, opened 2025-12-25
Implement clamping of estimates for segment times

### LINKED ISSUE #2638: Golf HCP split calculation appears to create humanly impossible splits
Not sure of exactly how the backend calculation works on this one but I do know that for user experience it shouldn't ever make a split that's better than your gold.

Here's an image comparing gold splits to HCP, as you can see it's creating some splits that are better than golds, and in some cases are in fact better than a human has ever performed.
Image

I would go so far as to say it should avoid passing your best split times too, if possible. Seems unreasonable for a handicap (ie, a moderately above average run) to expect you to play better than you've ever played up to that point. Left column is best split times and right is HCP, and HCP appears to take the lead in some places.
Image

A friend of mine speculated that this is being caused by skipping splits; ie if I skip a split, the handicap calculation counts that level as 0:00, and counts the next split as a much larger time than it should since it adds the two previous. I can't confirm if this is the issue but it sounds plausible so I thought I'd mention.

rodrigosiqueiradev:
This update introduces a new mechanism in the `HCPComparisonGenerator` class to dynamically scale segment times. A do-while loop has been added to iteratively adjust the scale factor based on clamping conditions, ensuring the total time aligns with a specified average. Additionally, the cumulative time calculation logic has been updated to support this new scaling approach, enhancing the accuracy and flexibility of segment time calculations. In other words, HCP will no longer estimate a time lower than best time for a segment. fixes #2638


wooferzfg:
Could you do some testing on the splits in #2638 and show the new results?

rodrigosiqueiradev:
Hmm, I'd very much like to have those splits to show an exact before/after. I'll try to prepare some scenario that I can take some screenshots to demonstrate it. For now all the testing I did locally and with some feedback from another runner indicated it's working nicely. But give me some time to try and prepare some splits to show it up.

wooferzfg:
It looks like for all but 1 of the segments, it's just using the best segment time now. That's really intended?

rodrigosiqueiradev:
The way that it was designed is that from the, let's say the default 20 last finished runs and 8 best of those, the average final time is calculated and the time per segment is estimated and distributed based on how much of the total time each segment takes in average, so HCP at a segment level is a secondary outcome of the comparison, while the primary goal is the total time.

A PB in itself is the best the runner ever got. In talks I've had with Smaugy (who gave the idea to implement this). The idea is to give a moving goal that considers only a few recent runs to ease up on the psychological impact of chasing only after PBs and move the objective into improving your consistency.

Yes, the Percentile comparison is a good solution, but I believe that it solves another scenario.

In the case we're looking at, the golds were from very early attempts, around the 200-300 mark if I recall correctly, so it can be that maybe strategy for the split changed, or some other factor like change in route and the gold is there in history until it's cleaned up manually. 

Perhaps it could be a clarity issue and documenting that HCP should primarily be used to look at the total time firstly and not directly comparable to best segments and in that case the PR should be discarded with no changes.

I'm very open to feedbacks and opinions on this overall, you guys created the whole thing after all :D 

I've taken another pics since I forgot to include the total time in the previous one...
splits

wooferzfg:
> time per segment is estimated and distributed based on how much of the total time each segment takes in average

Shouldn't you be distributing the time based on the variance? Is that what you're doing, or is it just based on the segment length?

rodrigosiqueiradev:
I can certainly investigate more. Two core points comes to mind, skipped splits, I need to revisit that case to check how it behaves. The other point is more conceptual, is that this comparison has a set window of data to work with, other comparisons use the entire dataset available if I'm not mistaken in their approaches, so I'd expect to see difference there.

The initial calculation, assuming no skipped splits, I believe are correct, but will check them again to be extra sure.

As for the scaling question above, it's within that range of attempts a given segment represents in average, for example, 10% of the total time of the run, that's where the scaling factor comes from when assigning a time to the split estimate (or more like guesstimate as it stands). 

I'll take a look into the other comments made with a fresh mind tomorrow.

rodrigosiqueiradev:
After some changes I got to a point where the numbers are making more sense. But I came across a problem that I'm not sure how to handle, regarding skipped splits, since when it's skipped time is flowing into the next valid segment. Don't know yet if there's a better way around this. Will dig some more into it. 

Also took into consideration the suggestion of using variance.

splits

PR #2679, opened 2026-04-05
Add per-component font override support

soarqin:
## Summary

- Add a unified per-component font override system that allows individual layout components to override the global Timer, Times, and Text fonts through the Layout Editor, without affecting other components.
- Add [GlobalFontConsumer] attribute to core base classes (InfoTextComponent, InfoTimeComponent) so the Layout Editor shows the correct font override rows.

### New files
- **FontOverrides** class (ICloneable, IDisposable) — holds three optional font overrides with apply/restore pattern for safe font swapping
- **GlobalFont** flags enum and **GlobalFontConsumerAttribute** — components declare which global fonts they consume
- **FontOverridePanel** UserControl — Layout Editor UI for per-component font override checkboxes and font choosers
- **FontOverridesMust** / **LayoutSerializationFontOverridesMust** — comprehensive xUnit tests

### Modified files
- **LayoutComponent** — extended with FontOverrides property and Clone() method
- **ComponentRenderer** — applies per-component font overrides around each component's Draw/Update call via 	ry/finally
- **Layout.Clone()** — deep-copies LayoutComponent wrappers with defensive pattern matching for plugin safety
- **XMLLayoutSaver** / **XMLLayoutFactory** — serialize/deserialize FontOverrides with backward compatibility (missing element = no overrides)
- **LayoutSettingsDialog** — snapshots FontOverrides on open, restores on cancel with proper Font disposal
- **LayoutEditorDialog** — double-click guard extended to also check [GlobalFontConsumer] attribute

## Design

The system works by temporarily swapping the global font in LayoutSettings before calling a component's Draw/Update, then restoring the original font in a finally block. Components don't need any code changes — they continue reading from state.LayoutSettings.TimerFont/TimesFont/TextFont as usual.

Components opt in by adding [GlobalFontConsumer(GlobalFont.XXX)] to their class. The attribute is Inherited = true, so derived classes automatically participate. Components that had their own font override settings can provide a MigrateFontOverrides(FontOverrides) method to populate the new unified FontOverrides on the LayoutComponent wrapper.

## Testing

- 485 tests pass (including new tests for FontOverrides clone/dispose/apply/restore and XML serialization round-trip)
- Build: 0 errors, 0 warnings

### DIFF
--- a/src/LiveSplit.Core/Options/FontOverrides.cs
+++ b/src/LiveSplit.Core/Options/FontOverrides.cs
@@ -1,0 +1,80 @@
+using System;
+using System.Drawing;
+
+namespace LiveSplit.Options;
+
+public class FontOverrides : ICloneable, IDisposable
+{
+    public bool OverrideTimerFont { get; set; }
+
+    public Font TimerFont { get; set; }
+
+    public bool OverrideTimesFont { get; set; }
+
+    public Font TimesFont { get; set; }
+
+    public bool OverrideTextFont { get; set; }
+
+    public Font TextFont { get; set; }
+
+    public bool HasOverrides => OverrideTimerFont || OverrideTimesFont || OverrideTextFont;
+
+    public void ApplyTo(LayoutSettings settings, out Font origTimer, out Font origTimes, out Font origText)
+    {
+        if (settings == null)
+        {
+            throw new ArgumentNullException(nameof(settings));
+        }
+
+        origTimer = settings.TimerFont;
+        origTimes = settings.TimesFont;
+        origText = settings.TextFont;
+
+        if (OverrideTimerFont && TimerFont != null)
+        {
+            settings.TimerFont = TimerFont;
+        }
+
+        if (OverrideTimesFont && TimesFont != null)
+        {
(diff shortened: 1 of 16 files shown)


soarqin:
> Making sure I understand the changes - why is `GlobalFontConsumer` being added to `InfoTextComponent` and `InfoTimeComponent` if the components aren't using it yet? What does this do for components that don't have the font overrides in the component settings yet?

1. If a Component does not use this attribute, the font override groupbox is hidden as well.
2. Components can mark their used fonts by adding `GlobalFontConsumer` attribute if they want font override groupbox shown, but are not enforced to do this. This is just an optional attribute.
3. The `GlobalFontConsumer` attribute is marked `Inherited = true`, so that any Component that inherits `InfoTextComponent` and `InfoTimeComponent` would inherit its font overridings' attribute.
Thus even if you don’t specify (or forget to specify) the attribute for these inherited components, the font override feature will still work properly. Of course, if your inherted implementation removed some use of kinda fonts, you can also force the attribute to be set to a new value in inherited Components.

soarqin:
PRs for all 16 font overridable submodules are created on their own repo.

PR #2706, opened 2026-06-01
[ColorPickerDialog] Improve keyboard usability

Kuhaku9893:
- Adjust tab stop targets and order
- Select all text in a NumericUpDown when focusing on it with the Tab key
- Suppress integer conversion to prevent the cursor from jumping back to the beginning while entering hexadecimal color values


Kuhaku9893:
I was trying to resolve a conflict but accidentally deleted the changes, so I'm resubmitting a pull request in #2711.

Issue #2716, opened 2026-07-18
Extra space in Times and Timer text rendering when using DPI Aware setting


Eiim:
When not using DPI Aware, digits are a reasonable distance apart:

<img width="600" height="1083" alt="Image" src="https://github.com/user-attachments/assets/efaf58d7-4f51-4f45-b28c-aa662b1d0573" />

When using DPI Aware, the space between digits in times and the timer (but not in regular text) is excessive:

<img width="533" height="922" alt="Image" src="https://github.com/user-attachments/assets/a386cab1-ada3-442a-b4c2-dcf3dae6c96b" />

This behavior seems to be independent of the font used.

PR #2717, opened 2026-07-18
Fix time spacing with high DPI (#2716)

Eiim:
SimpleLabel uses MeasureText from WinForms to measure the width of digits that it's rendering, but renders them with GDI+ GraphicsPath. These produce different results when the application is DPI-aware and DPI scaling is enabled in Windows. This PR manually scales the widths to match. The alternative seems to be to use g.MeasureCharacterRanges from GDI+, which would arguably mirror the text-rendering function more cleanly, but might affect non-DPI-aware rendering (it uses floats natively instead of ints, so I think there's potential rounding differences between them?), so I went with this approach to stay on the safe side. It also was relatively easy to implement given that I don't know much .NET at all.

### DIFF
--- a/src/LiveSplit.Core/UI/SimpleLabel.cs
+++ b/src/LiveSplit.Core/UI/SimpleLabel.cs
@@ -105,7 +105,7 @@ public class SimpleLabel
                 LineAlignment = VerticalAlignment
             };
 
-            int measurement = MeasureText(g, "0", Font, new Size((int)(Width + 0.5f), (int)(Height + 0.5f)), TextFormatFlags.NoPadding).Width;
+            int measurement = MeasureCharacterActualWidth("0", g);
             float offset = Width;
             int charIndex = 0;
             SetActualWidth(g);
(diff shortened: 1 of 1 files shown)


src/LiveSplit.Core/UI/SimpleLabel.cs
```diff
@@ -231,14 +231,28 @@ private float MeasureActualWidth(string text, Graphics g)
...
+    
+    // Intended to measure the width of a single character, taking into account DPI scaling
+    // For longer strings, MeasureActualWidth should be used instead to keep digit spacing consistent
+    private int MeasureCharacterActualWidth(string text, Graphics g)
+    {
+        int width = MeasureText(g, text, Font, new Size((int)(Width + 0.5f), (int)(Height + 0.5f)), TextFormatFlags.NoPadding).Width;
+        
+		if (Font.Unit != GraphicsUnit.Point)
+        {
+            width = (int)((width * 96f / g.DpiY) + 0.5f);
```
wooferzfg, line 251: Where is this `96f` magic number coming from?

Eiim: 96 is the default DPI in Windows, see https://learn.microsoft.com/en-us/windows/win32/learnwin32/dpi-and-device-independent-pixels. Looks like GDI has a constant for it, so you could replace it with that constant instead. If you're doing that, I would also make the 72 PPI value used elsewhere in this file a constant.

wooferzfg: Feel free to make that change

PR #2723, opened 2026-07-26
Support Native Subsplits

CryZe:
Update livesplit-core together with the generated C# bindings and native libraries to a revision with native segment groups.

Translate those groups back into LiveSplit's legacy dash- and brace-based names when importing runs. This keeps existing subsplit handling working without requiring broader changes throughout LiveSplit.

Cover both native group conversion and legacy files upgraded internally by livesplit-core.


src/LiveSplit.Core/Model/RunFactories/StandardFormatsRunFactory.cs
```diff
@@ -94,6 +94,40 @@ private static Image ParseImage(IntPtr imagePtr, ulong length)
...
+    {
+        int segmentCount = checked((int)run.Len());
+        var names = new string[segmentCount];
+        for (int i = 0; i < segmentCount; i++)
+        {
+            names[i] = run.Segment((ulong)i).Name();
+        }
+
+        // The C API does not expose segment groups directly, so use the
+        // canonical LiveSplit serialization as the compatibility bridge.
```
wooferzfg: Why can't the C API expose these directly?

CryZe: Oh, apparently I never pushed the actual commit. This is an outdated one, the C API is updated and used in the new commit.

wooferzfg (approved):

CryZe:
Before you merge it or something, I wonder if we should reduce the size of the DLLs. Like, these DLLs include the entirety of livesplit-core, and we really only use it for parsing. And actually, I'm thinking the same thing for LiveSplit One, where in the web I always bundle way more than necessary, so I wonder if I should investigate that first, so we can just have like a parsing-only build, so these DLLs are a lot smaller.

wooferzfg:
That sounds like a good idea, but probably doesn't have to block this PR

CryZe:
Alright, I put up an initial subset DLL, but I'd like to review and test it further (and also first merge the subsetting into livesplit-core).

PR #2734, opened 2026-09-01
Add AutoSplittingRuntime child support

ransagy:
Follow-up to LiveSplit/LiveSplit.AutoSplitters#2878 and Dandara #2516. Allows a single Game entry to carry an optional <AutoSplittingRuntime> child with WASM URL for LiveSplit One/druid/OBS, while keeping the primary Component/ASL entry for desktop. Old builds ignore the child. Based on the suggested format in the first linked PR.


ransagy:
Now that the format supports a dedicated AutoSplittingRuntime for the core based clients, the question is whether Desktop should show/expose both splitters for a game, or just keep the Component one as-is.

If yes, it'd mean a small addition to the auto-splitter panel (a picker or toggle when multiple splitters exist for a game, showing each one's description/author so the user picks based on that).

In order for the rest of the behavior to stay the same activate, deactivate, settings, website, .lss round-trip, speedrun.com integration — the `AutoSplitters` property type would need to change from one entry per game to a list per game, which likely breaks the API for anything directly accessing the dictionary.

@CryZe @AlexKnauth @wooferzfg happy to hear your thoughts on what you think makes the most sense per LiveSplit's design and goals.

test/LiveSplit.Tests/AutoSplitterTests/AutoSplitterXML.cs
```diff
@@ -62,5 +62,19 @@ public void TestAutoSplittersXML()
             "Auto Splitters need to have valid URLs");
         Assert.True(!autoSplitters.Values.Any(x => x.URLs.Any(y => Regex.IsMatch(y, "https://github.com/[^/]*/[^/]*/blob/"))),
             "URLs leading to GitHub should use the raw file link");
+
+        // Validate AutoSplittingRuntime child if present
```
CryZe: These checks only run for entries where `AutoSplittingRuntime` is already not null. If `CreateFromXmlElement` stops reading the new child, all of the checks still pass. The test also skips this part when the local XML file is older and has no runtime entry.

Could we add a small fixed XML example and check the parsed values directly? The suggested code also checks that the ASL remains the main splitter and that `Clone()` copies the runtime information.

```suggestion
        // Test parsing with a fixed example so this does not depend on the downloaded XML.
        var runtimeDocument = new XmlDocument();
        runtimeDocument.LoadXml("""
            <AutoSplitter>
                <Games>
                    <Game>Example Game</Game>
                </Games>
                <URLs>
                    <URL>https://example.com/example.asl</URL>
                </URLs>
                <Type>Script</Type>
                <Description>ASL splitter</Description>
                <Website>https://example.com/asl</Website>
                <AutoSplittingRuntime>
                    <URL>https://example.com/example.wasm</URL>
                    <Description>WASM splitter</Description>
                    <Website>https://example.com/wasm</Website>
                </AutoSplittingRuntime>
            </AutoSplitter>
            """);

        AutoSplitter runtimeAutoSplitter = AutoSplitterFactory.CreateFromXmlElement(runtimeDocument.DocumentElement);

        Assert.Equal(AutoSplitterType.Script, runtimeAutoSplitter.Type);
        Assert.Equal("https://example.com/example.asl", runtimeAutoSplitter.URLs.Single());
        Assert.NotNull(runtimeAutoSplitter.AutoSplittingRuntime);
        Assert.Equal("https://example.com/example.wasm", runtimeAutoSplitter.AutoSplittingRuntime.URL);
        Assert.Equal("WASM splitter", runtimeAutoSplitter.AutoSplittingRuntime.Description);
        Assert.Equal("https://example.com/wasm", runtimeAutoSplitter.AutoSplittingRuntime.Website);

        AutoSplitter runtimeClone = runtimeAutoSplitter.Clone();
        Assert.NotSame(runtimeAutoSplitter.AutoSplittingRuntime, runtimeClone.AutoSplittingRuntime);
        Assert.Equal(runtimeAutoSplitter.AutoSplittingRuntime.URL, runtimeClone.AutoSplittingRuntime.URL);
        Assert.Equal(runtimeAutoSplitter.AutoSplittingRuntime.Description, runtimeClone.AutoSplittingRuntime.Description);
        Assert.Equal(runtimeAutoSplitter.AutoSplittingRuntime.Website, runtimeClone.AutoSplittingRuntime.Website);

        // Validate AutoSplittingRuntime children in the downloaded XML.
```

CryZe (changes requested):
The XML format and parser look good. I would like one test change before this is merged. The inline suggestion explains the issue and includes code that should be directly applicable.

ransagy:
Applied; looks good logic/syntax wise.
I can't quite run the test locally at the moment (no available Win machine here) so waiting for CI to do the final verification.

